PR #2, opened 2017-03-28
Calculate likelihoods for split alignments using pseudo-haplotypes.

johanneskoester:
Further, this PR refines the calculation of SNV likelihoods.


johanneskoester:
@dlaehnemann this might affect you as well, because it entails quite some refactoring.

PR #3, opened 2017-04-05
SingleCellBulkModel

dlaehnemann:
This branch, rebased on the most recent master branch commit, implements a SingleCellBulkModel, where the single cell amplification bias model is based on Lodato et al., 2015, Science (doi:10.1126/science.aab1785). It includes tests and a basic one has also been added to the TumorNormalModel along the way.


src/model/priors/single_cell_bulk.rs
```diff
@@ -0,0 +1,578 @@
...
+                    }).collect_vec());
+                NotNaN::new(*p_single).expect("posterior probability is NaN")
+            }
+        ).into_option().expect("prior has empty allele frequency spectrum");
+
+        let k_bulk = 0..n_obs_bulk + 1;
+        let afs_bulk = k_bulk.map(|k| AlleleFreq(k as f64/n_obs_bulk as f64));
+        let (_, map_bulk) = afs_bulk.minmax_by_key(
+            |af_bulk| {
+                let p_bulk = likelihood_bulk(*af_bulk, None);
```
johanneskoester: No prior here?

src/model/priors/single_cell_bulk.rs
```diff
@@ -0,0 +1,578 @@
...
+        }
+    }
+}
+
+
+impl PairModel<DiscreteAlleleFreqs, ContinuousAlleleFreqs> for SingleCellBulkModel {
+
+    fn prior_prob(&self, _: AlleleFreq, _: AlleleFreq, _: Variant) -> LogProb {
+        // TODO: stick in the InfiniteSitesNeutralVariationModel here?
+        LogProb::ln_one()
```
johanneskoester: Are you sure that this is not applicable to return prob_rho for a given allele frequency combination?

dlaehnemann: You mean sticking prob_rho here as a prior for the single cell sample? I guess that's somewhat like what I originally thought, but then abandoned, as we were unsure about how to view this. But looking at the implementation now, this does make total sense, again... ;)
Also, it might make the implementation of alternative amplification bias models cleaner.

And for the bulk sample we should probably also adjust for the level of heterozygosity. But this would need to also account for the somatic mutation rate, so I guess it couldn't be the plain InfiniteSitesNeutralVariation model.

src/model/priors/single_cell_bulk.rs
dlaehnemann: Oops, that one ... got lost ...

Well, as the prior is currently flatly set to 1, I guess I just completely forget about it. Will of course add it back in, also for the bulk in fn joint_prob().

src/model/priors/single_cell_bulk.rs
johanneskoester: Yes indeed, for bulk it is also not yet clear for me how to have a good prior. Uniform is at least a start.

dlaehnemann: I would then go for a signature of:

```
fn prior_prob(&self, af_single: Option<AlleleFreq>, af_bulk: AlleFreq, variant: Variant) -> LogProb
```

I'll need separate priors for single cell and bulk sample and thus prior_prob needs to return something different for the single cell than for the bulk sample. This way, I could make its return value dependent on whether I pass in a af_single or not. However, this requires some code refactoring, as this goes back to the function signature in pub trait PairModel in model/priors/mod.rs

dlaehnemann: And we'll have to get n_obs_single and k_s into the prior_prob signature, as we don't have a single prior per possible underlying single cell allele frequency, but a whole distribution dependent on these two values. So maybe this is not the classical intention for this prior here, after all?

johanneskoester: Ok, I see. better leave it like it is now, then. I am still not sure why you would need the option type above. If you don't have a single cell sample, there is no need to have a PairModel, isn't it? Also, can't the prior just return the combined prior for both single and bulk?

dlaehnemann: Well, in effect, the probabilities for the single cell sample and the bulk sample are currently independent (as are the maximum a posteriori allele frequency estimates), so theoretically a `pub trait Model` for just the single cell sample might do for variant calling in the single cell sample. The bulk so far is only needed for somatic vs. germline classifications in prosolo.

But eventually, the bulk could also provide some kind of prior information for the single cell sample, thus linking the two again.

But as for prob_rho and being a prior: It is prior knowledge that we have, but not on the plain probability of an underlying allele frequency. Rather, it is a full-blown probability mass function of seeing a certain amount `k` of alternative alleles out of the local coverage `n` for a possible underlying allele frequency. A prior would instead just weigh between the possible underlying allele frequencies, right?

For now, I'll add the `panic`.

johanneskoester: That is right. So, right now there is no prior at all for the underlying allele frequencies? Then, it makes sense to not implement the prior_prob.

dlaehnemann: Well, I guess the bulk prior would be probability density function based on heterozygosity plus some adjustment for a somatic mutation rate, as all alllele frequencies [0,1] are possible.

And the single cell prior would need to be an analogous probability mass function, if considered alone. But it could instead be based on whatever we see in the bulk, when that is available -- as an example, if we see evidence for a somatic mutation from heterozygous to homozygous alt in the bulk, this should make seeing such a homozygous alt genotype in a single cell much more likeliy. Am I making sense?

johanneskoester: Yes, I agree.

johanneskoester:
I am fine with merging now.

dlaehnemann:
Thanks a lot for the code review and the useful discussion!

PR #4, opened 2017-04-20
2 bugfixes for SNV handling + workaround for non-inclusive range starts/inclusive range ends

dlaehnemann:
* non-inclusive range starts and inclusive range ends not in stable Rust, workaround with offsets
* bugfix: include all covering reads for SNVs, i.e. also those that cover the SNV with the first read position
* bugfix: use correct covering position within reads, not always the first of the respective CIGAR operation

### DIFF
--- a/src/model/priors/single_cell_bulk.rs
+++ b/src/model/priors/single_cell_bulk.rs
@@ -29,7 +29,7 @@ impl SingleCellBulkModel {
         let allele_freqs = (0..ploidy + 1).map(|m| AlleleFreq(m as f64 / ploidy as f64)).collect_vec();
         SingleCellBulkModel {
             allele_freqs_single: allele_freqs,
-            allele_freqs_bulk: AlleleFreq(0.0)..AlleleFreq(1.0)
+            allele_freqs_bulk: AlleleFreq(0.0)..AlleleFreq(1.00000001) // TODO: make use of inclusive ranges as soon as they arrive in stable rust
         }
     }
 
@@ -142,14 +142,12 @@ impl PairModel<DiscreteAlleleFreqs, ContinuousAlleleFreqs> for SingleCellBulkMod
         };
         let k_single = 0..n_obs_s + 1;
 
-        let k_start = if *af_bulk.start == 0.0 { // 0 as a bulk range start is always inclusive
-            0 as u64
-        } else {
-            // any non-zero bulk range start is always exclusive
-            (*af_bulk.start * n_obs_bulk as f64).floor() as u64 + 1
-        };
-        // any bulk range end is always inclusive
-        let k_end = (*af_bulk.end * n_obs_bulk as f64).floor() as u64 + 1;
+        // stable rust ranges currently only allow inclusive start points, for exclusive add
+        // 0.00000001 in the Event definition
+        let k_start = (*af_bulk.start * n_obs_bulk as f64).ceil() as u64;
+        // stable rust ranges currently only allow exclusive end points, for including an end point,
+        // add 0.00000001 in the Event definition
+        let k_end = (*af_bulk.end * n_obs_bulk as f64).floor() as u64;
         let k_bulk = k_start..k_end;
 
         // sum up all possible discrete bulk allele frequencies with current number of observations
@@ -355,7 +353,7 @@ mod tests {
 
         // single cell is het against het germline
         let af_single = vec![AlleleFreq(0.5)];
-        let af_bulk = AlleleFreq(0.25)..AlleleFreq(0.75);
+        let af_bulk = AlleleFreq(0.25000001)..AlleleFreq(0.75000001);
 
         let variant = Variant::SNV(b'T');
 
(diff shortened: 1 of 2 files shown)


src/model/sample.rs
```diff
@@ -47,7 +47,8 @@ pub fn prob_read_snv(record: &bam::Record, cigar: &[Cigar], start: u32, variant:
         for c in cigar {
             match c {
                 // potential SNV evidence
-                &Cigar::Match(l) | &Cigar::Diff(l) if contains_start(pos, l) => {
+                &Cigar::Match(l) | &Cigar::Diff(l) | &Cigar::Equal(l) if contains_start(pos, l) => {
+                    qpos = start - pos as u32;
```
johanneskoester: Good catch!

src/model/priors/single_cell_bulk.rs
```diff
@@ -29,7 +29,7 @@ impl SingleCellBulkModel {
         let allele_freqs = (0..ploidy + 1).map(|m| AlleleFreq(m as f64 / ploidy as f64)).collect_vec();
         SingleCellBulkModel {
             allele_freqs_single: allele_freqs,
-            allele_freqs_bulk: AlleleFreq(0.0)..AlleleFreq(1.0)
+            allele_freqs_bulk: AlleleFreq(0.0)..AlleleFreq(1.00000001) // TODO: make use of inclusive ranges as soon as they arrive in stable rust
```
johanneskoester: I believe this is not necessary. We can interpret the endpoint of the range as we want, since we never iterate over these allelefreq ranges.

src/model/priors/single_cell_bulk.rs
```diff
@@ -142,14 +142,12 @@ impl PairModel<DiscreteAlleleFreqs, ContinuousAlleleFreqs> for SingleCellBulkMod
...
-            (*af_bulk.start * n_obs_bulk as f64).floor() as u64 + 1
-        };
-        // any bulk range end is always inclusive
-        let k_end = (*af_bulk.end * n_obs_bulk as f64).floor() as u64 + 1;
+        // stable rust ranges currently only allow inclusive start points, for exclusive add
+        // 0.00000001 in the Event definition
+        let k_start = (*af_bulk.start * n_obs_bulk as f64).ceil() as u64;
+        // stable rust ranges currently only allow exclusive end points, for including an end point,
+        // add 0.00000001 in the Event definition
+        let k_end = (*af_bulk.end * n_obs_bulk as f64).floor() as u64;
```
johanneskoester: I vote for keeping the allelefreq ranges clean and handle the offset just here.

src/model/sample.rs
dlaehnemann: Well, my simulated ProSolo testing data gave ... unexpected results. The observations flag and the debug output were of great help!

src/model/priors/single_cell_bulk.rs
dlaehnemann: If I follow the calling logic correctly, marginal_prob() does take this range and passes it into a call of joint_prob() over this range, so I think we do need to set the offset here. But please correct me, if I'm missing something.

src/model/priors/single_cell_bulk.rs
dlaehnemann: The trouble is, that ProSolo defines both inclusive AND exclusive start points, and both inclusive AND exclusive end points for ranges (starting here: https://github.com/PROSIC/prosolo/blob/master/src/call.rs#L81). Thus until we have ranges that can pass this information on to this point, I don't see how I could handle this here. But I do agree that the current solution isn't pretty and I'd prefer treating this here!

And in any case, thanks for the thorough code review!

src/model/priors/single_cell_bulk.rs
johanneskoester: It does, but it is up the the implementation of joint_prob to interpret the upper bound. At least my implementations treat it as inclusive. We should definitely add a TODO though that normal Range shall be replaced by the inclusive range type once it has been stabilized.

dlaehnemann: Is the current TODO too ambiguous, or do you mean the implementation of joint_prob() below or the respective Event definitions in ProSolo?

src/model/priors/single_cell_bulk.rs
johanneskoester: Ok, I understand the issue now. Would the following work?

```
/// An allele frequency range (left- and right-inclusive by default).
pub struct AFRange {
    inner: pub Range<AlleleFreq>,
    pub left_exclusive: bool,
    pub right_exclusive: bool
}

impl {
    pub fn left_exclusive(self) -> Self {
        self.left_exclusive = true;
    }
    
    pub fn right_exclusive(self) -> Self {
        self.left_exclusive = true;
    }
}
```

An event could then be instantiated like this:
```
libprosic::call::pairwise::PairEvent {
            name: "germline_hom_ref".to_owned(),
            af_case: vec![AlleleFreq(0.0)],
            af_control: AFRange(AlleleFreq(0.0)..AlleleFreq(0.25)).right_exclusive()
}
```

dlaehnemann: Sure, that looks like a viable solution. I also considered implementing something, but then figured it would be worth waiting for std::ops::RangeInclusive to become stable for a proper solution and instead going for the easiest hack in the meantime. But then I realise, that the new std::ops::Range also doesn't account for an exclusive start, yet. So we would have to still work around this in the future -- maybe with something like `std::ops::Range.skip_while(|x| x == start)` at range definition, although I'm not sure about types (range vs. resulting iterator). So maybe we need to implement this ourselves, after all.

Currently I'm doing some analysis with this state of the single-cell-bulk branch and current ProSolo, so I'd leave this open for now and address it in the coming weeks.

johanneskoester: Ok

dlaehnemann:
Closing for now, will reopen once the ranges and everything else are cleaned up, so that there aren't continuous eMail-updates before it's worth looking at again. Thanks for the feedback so far.

johanneskoester:
Ok, you're welcome!

PR #5, opened 2017-05-04
Constrained sampling of read overlaps

johanneskoester:
For indels, we now sample reads that overlap the start or end of the variant by a configurable maximum. This ensures that alt and ref reads are sampled without bias.
Otherwise, the bias is as follows. If the overlap is too large, an alt read is (a) softclipped but with reduced MAPQ, (b) hardclipped such that the alt sequence is lost, (c) not mapped at all. All three cases are unfair compared to the ref reads and lead to biased allele frequency estimates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "GPL-3.0"
 [dependencies]
 bio = "0.10.*"
 log = "0.3.*"
-rust-htslib = "0.9.*"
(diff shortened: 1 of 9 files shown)


johanneskoester:
@dlaehnemann this is a major change in the sample class and its parameters. In principle, it should not really affect you apart from the need to adapt the instantiation of the struct in prosolo.

PR #6, opened 2017-05-09
`SingleCellBulkModel`: major bugfixes, impacting core library

dlaehnemann:
Here comes the major cleanup of the `SingleCellBulkModel` debugging, including some changes to core parts of the library. So a meticulous code review would be very welcome!

1. bugfix: include all covering reads for SNVs, i.e. also those that cover the SNV with the first read position, which were previously omitted
2. bugfix: use correct covering position within reads, not always the first of the respective CIGAR operation. This and the following bugfix led to a major overhaul of CIGAR string parsing in `prob_read_snv()` (tests included)
3. bugfix: do not push reads that have the reference position of an SNV skipped or deleted into `observations` (tests included). This required that `read_observation()` return an Option wrapped result: `Some(prob_ref, prob_alt)`.
4. bugfix: discretisation of bulk ranges in `SingleCellBulkModel`'s `joint_prob()` was incorrect, is fixed now (tests included). After implementing tests, I eventually included a modified version of the proper `AFRange` to allow for inclusive/exclusive range start and end points that @johanneskoester suggested in the previous pull request. Some range instantiations in the src/model/mod.rs tests need review, they are marked with TODO.
5. bugfix: previously, likelihoods were calculated even for candidate variants where one of the sample pileups does not have any observations/reads covering it. This lead to divisions by zero in SingleCellBulkModel's joint_prob() and map(), and thus to program crashes of ProSolo. Now, no likelihoods are calculated if one of the pileups is empty -- instead, a missing value (None) is returned.


johanneskoester:
Thanks a lot! Can you point me to the code lines related to (5)?

dlaehnemann:
For the check for empty pileups, the changesare in src/call/pairwise.rs in previous line 192.

The zero divisions would happen in src/model/priors/single_cell_bulk.rs in the divisions by n_obs_bulk and/or n_obs_s, in lines 160, 168, 225, and 233.

I hope this gives you what you were looking for?!

src/call/pairwise.rs
```diff
@@ -189,7 +189,15 @@ pub fn call<A, B, P, M, R, W, X, F>(
...
-                        Some(pileup.posterior_prob(&event.af_case, &event.af_control))
+                        // if either of the pileups is empty, no valid probability could be
+                        // calculated for it -- and in SingleCellBulkModel's joint_prob() and map()
+                        // this would result in a division by zero and thus a program crash
+                        if  !pileup.case_observations().is_empty() &&
+                            !pileup.control_observations().is_empty() {
+                                Some(pileup.posterior_prob(&event.af_case, &event.af_control))
+                            } else {
+                                None
+                            }
```
johanneskoester: Actually this is not the case in general. Even if both pileups are empty, you should still get a posterior probability. This probability should simply reflect the prior assumptions. E.g. with flat priors, it should be proportional to the fraction an event takes in the whole event space, with real informative priors, it should be equivalent to the prior expected heterozygosity or mutation rate. Further, if only one pileup is empty the prior for the empty pileup can still help to lead to a more accurate posterior. Imagine a tumor-normal sample pair where I have a good coverage in the tumor, saying that there is a variant at some allele frequency. Even if I have no coverage in the normal, my prior tells me that there is some likelihood of having the same variant in the normal sample. Hence it will nicely adjust my posterior probability.

So my feeling is that if you get a division by zero error in your model, there must be something wrong in there. Maybe you can handle this case in a special way?

src/model/mod.rs
```diff
@@ -18,7 +18,7 @@ use model::sample::{Sample, Observation};
 
 pub type AlleleFreq = NotNaN<f64>;
 pub type DiscreteAlleleFreqs = Vec<AlleleFreq>;
-pub type ContinuousAlleleFreqs = Range<AlleleFreq>;
+pub type ContinuousAlleleFreqs = AFRange;
```
johanneskoester: I think we have redundant type names now. Maybe simply call `AFRange` `ContinuousAlleleFreqs` instead?

src/model/mod.rs
```diff
@@ -435,8 +481,8 @@ mod tests {
     #[test]
     fn test_same_pileup() {
         let variant = Variant::Deletion(3);
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.0)..AlleleFreq(1.0);
+        let tumor_all = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) );
+        let tumor_alt = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: When reading this, I think the API can be even more improved. Having AFRange and AlleleFreq in the invocation is somehow redundant. Would be better to allow this:
```
AFRange::inclusive(0.0..1.0);
```

src/model/mod.rs
```diff
@@ -31,6 +31,51 @@ pub trait AlleleFreqs: Debug {}
 impl AlleleFreqs for DiscreteAlleleFreqs {}
 impl AlleleFreqs for ContinuousAlleleFreqs {}
 
+/// An allele frequency range
+#[derive(Debug)]
+pub struct AFRange {
+    pub inner: Range<AlleleFreq>,
+    pub left_exclusive: bool,
+    pub right_exclusive: bool
+}
```
johanneskoester: Please remove the `pub` from inner and implement the Deref trait for AFRange. This way, we can directly iterate over AFRange and also get access to start and end.
Simply add
```
impl Deref for AFRange {
    type Target = Range<AlleleFreq>;

    fn deref(&self) -> &Range<AlleleFreq> {
        &self.inner
    }
}
```

src/model/mod.rs
```diff
@@ -474,8 +520,8 @@ mod tests {
     #[test]
     fn test_empty_control_pileup() {
         let variant = Variant::Deletion(3);
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.0)..AlleleFreq(1.0);
+        let tumor_all = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) );
+        let tumor_alt = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: Yes. Numerically it does not matter in my case, but now that we are able to express this, it should be done.

src/model/mod.rs
```diff
@@ -514,8 +560,8 @@ mod tests {
     #[test]
     fn test_subclonal() {
         let variant = Variant::Deletion(3);
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.0)..AlleleFreq(1.0);
+        let tumor_all = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) );
+        let tumor_alt = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: Yes.

src/model/mod.rs
```diff
@@ -560,9 +606,9 @@ mod tests {
     #[test]
     fn test_absent() {
         let variant = Variant::Deletion(3);
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_ref = AlleleFreq(0.0)..AlleleFreq(0.0);
+        let tumor_all = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) );
+        let tumor_alt = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: Yes.

src/model/mod.rs
```diff
@@ -637,9 +683,9 @@ mod tests {
             prior_model
         );
 
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.05)..AlleleFreq(1.0);
-        let tumor_ref = AlleleFreq(0.0)..AlleleFreq(0.001);
+        let tumor_all = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(1.0) );
+        let tumor_alt = AFRange::inclusive( AlleleFreq(0.05)..AlleleFreq(1.0) ); // TODO: should this be left_exclusive() instead?
+        let tumor_ref = AFRange::inclusive( AlleleFreq(0.0)..AlleleFreq(0.001) ); // TODO: should this be (0.0)..(0.0) instead?
```
johanneskoester: I will think about this test, thanks for adding the TODOs.

src/model/priors/flat.rs
```diff
@@ -155,10 +155,10 @@ impl PairModel<ContinuousAlleleFreqs, DiscreteAlleleFreqs> for FlatTumorNormalMo
                 likelihood_tumor(af_tumor, Some(af_normal))
             };
 
-            let p_tumor = if af_tumor.start == af_tumor.end {
-                density(*af_tumor.start)
+            let p_tumor = if af_tumor.inner.start == af_tumor.inner.end {
```
johanneskoester: All these `*.inner.*` changes will become superfluous with the Deref trait implementation shown above.

src/model/sample.rs
```diff
@@ -36,40 +36,73 @@ pub fn prob_read_base(read_base: u8, ref_base: u8, base_qual: u8) -> LogProb {
...
+                    c += 1;
+                },
+                &Cigar::HardClip(_) => return None,
+                &Cigar::Back(_) => panic!("Back: Unsupported Cigar operation (not in SAMv1 spec)."),
+                &Cigar::Pad(_) => panic!("Pad: Unsupported Cigar operation (unclear SAMv1 spec).")
             }
         }
-        panic!("read does not enclose SNV");
+        // if no SNV evidence is found before passing the vpos, must be
+        return None;
```
johanneskoester: I don't yet understand why the loop is changed like this. Can you point me to the problems with the previous loop?

src/model/sample.rs
```diff
@@ -580,11 +612,16 @@ impl Sample {
                     let end_pos = record.end_pos(&cigar) as u32;
 
                     if pos <= start && end_pos >= start {
-                        let cigar = record.cigar();
-                        observations.push(
-                            self.read_observation(&record, &cigar, start, variant, chrom_seq)
-                        );
-                        n_overlap += 1;
+                        let obs = self.read_observation(&record, &cigar, start, variant, chrom_seq);
+                        match obs {
```
johanneskoester: Please use

```
if let Some(obs) = self.read_observation(...) {
    observations.push(o);
    n_overlap += 1;
}
```

src/model/sample.rs
```diff
@@ -619,10 +656,14 @@ impl Sample {
                     // read evidence
                     if overlap > 0 {
                         if overlap <= self.max_indel_overlap {
-                            observations.push(
-                                self.read_observation(&record, &cigar, start, variant, chrom_seq)
-                            );
-                            n_overlap += 1;
+                            let obs = self.read_observation(&record, &cigar, start, variant, chrom_seq);
+                            match obs {
```
johanneskoester: Same as above.

src/model/sample.rs
```diff
@@ -682,25 +723,31 @@ impl Sample {
...
             }
         };
 
-        Observation {
-            prob_mapping: prob_mapping,
-            prob_alt: prob_alt,
-            prob_ref: prob_ref,
-            prob_mismapped: LogProb::ln_one(), // if the read is mismapped, we assume sampling probability 1.0
-            evidence: Evidence::from(cigar)
+        match probs {
```
johanneskoester: `if let` like shown above.

src/model/sample.rs
```diff
@@ -981,9 +1029,14 @@ mod tests {
             let record = record.unwrap();
             let cigar = record.cigar();
             let obs = sample.read_observation(&record, &cigar, varpos, variant, &ref_seq);
-            assert_relative_eq!(*obs.prob_alt, *true_alt_prob, epsilon=0.001);
-            assert_relative_eq!(*obs.prob_mapping, *(LogProb::from(PHREDProb(60.0)).ln_one_minus_exp()));
-            assert_relative_eq!(*obs.prob_mismapped, *LogProb::ln_one());
+            match obs {
```
johanneskoester: `it let`

src/model/sample.rs
```diff
@@ -1089,4 +1142,89 @@ mod tests {
             assert_relative_eq!(*prob_alt, probs_alt[i], epsilon=0.001);
         }
     }
+
+    #[test]
+    fn test_prob_read_snv() {
```
johanneskoester: Thanks for the clean test case!

src/call/pairwise.rs
dlaehnemann: You are absolutely right.

So the clean solution is to include proper priors (one each for the bulk and the single cell sample, as these are distinct in the SingleCellBulkModel) and then handle the case that either `n_obs_s` or `n_obs_bulk` is zero separately. For the single cell sample, this is straightforward, as we are always testing across the cases 0, 0.5 and 1, and we then just return the priors for those. For the bulk sample this is more complicated, as we discretise to the possible allele frequencies, given the total coverage `n_obs_bulk`. But then this approach doesn't make sense for low total coverages, as the Event space with its ranges will be more finely resolved than the allele frequencies we look at. So maybe we should go back to a fixed granularity, as you have it for the TumorNormalModel, or at least have a minimum (and maybe maximum) resolution, e.g. at least 11 different allele frequencies from 0.0..1.0. Any thoughts? Any numerical insights?

src/model/sample.rs
dlaehnemann: There was the following major problem: It accepted `Del`s and `RefSkip`s before a potential `SoftClip` (something that conflicts with the SAMv1 spec) or before any reference match appearing in `seq` (which doesn't make any sense). Thus we first have to find the first Cigar operation that refers to `record.seq()[0]` at `record.pos()`, which is any operation consuming `record.seq()` positions. Only once we have this starting point, can we trust and adjust for following `Del`s and `RefSkip`s by increasing rpos and only then can we take a HardClip as a read end and return `None`. Theoretically, if we wanted to really enforce the SAMv1 spec, we would also have to check that no operation comes between a `SoftClip` and the respective read end or `HardClip` operation... 

Also, if you feel like comparing this to the Cigar string parsing logic of htslib, have a look at resolve_cigar2 in sam.c on github.com/samtools/htslib/sam.c (<https://github.com/samtools/htslib/blob/develop/sam.c#L1861>). ;)

In addition, there were these minor problems:
* the `SoftClip` operation wasn't accounted for, previously
* the `Back` operation doesn't exist in the SAMv1 spec, why is it in `rust-htslib`?
* the `Pad` operation doesn't have a clear explanation in the SAMv1 spec, so I wouldn't use it, as I wouldn't know how
* the loop couldn't return `None`, if the reference base of the SNV was actually deleted or skipped in the record
* qpos wasn't adjusted for the actual position within the current Cigar operation, but was always the first position of this Cigar operation

I hope, I have recapped all the things that have changed here. But I wouldn't be surprised to find further bugs / corner cases here in the future, especially with data from the wild...

src/call/pairwise.rs
johanneskoester: I think I am not enough into the model to answer this properly. If I understood correctly, your current prior is flat, right? So it would be reasonable to ensure that uniform posteriors are returned in the case of empty pileups. For more advice I need to understand the details. If you want, we can have a video conference later today or on Monday?

src/model/sample.rs
```diff
@@ -36,40 +36,73 @@ pub fn prob_read_base(read_base: u8, ref_base: u8, base_qual: u8) -> LogProb {
...
+            match &cigar[i] {
+                &Cigar::Match(_) |
+                &Cigar::Diff(_)  |
+                &Cigar::Equal(_) |
+                &Cigar::Ins(_)   | // this is unexpected, but bwa + GATK indel realignment can produce insertions before matching positions
+                &Cigar::SoftClip(_) => {
+                    c = i;
+                    break;
+                },
+                &Cigar::Del(_) => panic!("Cigar String: Unexpected Deletion operation found before any reference match."),
```
johanneskoester: Instead of panicking, it would be better to return `Result<(LogProb, LogProb), CigarError>` and define `CigarError` similar to `RecordBufferError` in this file, but with one variant per error type, i.e.:

```
quick_error! {
    #[derive(Debug)]
    pub enum CigarError {
        UnsupportedOperation(msg: String) {
            description("unsupported CIGAR operation")
            display("{} ", msg)
        },
        UnexpectedOperation(msg: String) {
            description("unexpected operation")
            display("{}", msg)
        },
        ...
    }
}
```

src/model/sample.rs
johanneskoester: Ok, makes sense to me, thanks! I think `Back` comes from the htslib source. I wanted to implement everything mentioned there. Maybe it never made into the SAM spec. I have just added one minor additional comment.

dlaehnemann: You are right, it is in the source code -- amazing... It's defined in `sam.h`, alongside all the SAMv1 spec Cigar operations and then it is used exactly three times, all within the function `bam_remove_B()` whose sole purpose seems to be to get rid of such Cigar operations with `samtools view -B`, should the input have it. Previously, it also used to be in the `bam_plcmd.c`, but is considered obsolete there since this commit from 4 years ago:
https://github.com/samtools/samtools/commit/e7ae7f96c7e78a2dd6eabdaed57037c483951929

So I'm not sure how important it is, to implement this one -- but maybe for completeness we should, and then I would opt for `panic!`/`quick_error!` handling that points out the obsoloteness.

src/call/pairwise.rs
dlaehnemann: Sure, I'm flexible this afternoon, just drop me a quick eMail of how and when.

src/model/sample.rs
dlaehnemann: Should we implement these `CigarError`s into rust-htslib, as they should be usable in any Cigar parsing that uses rust-htslib?

johanneskoester: For now let's leave it in here. We can always do that later, but since there is currently no functionality within Rust-Htslib to use it, I don't think it would be the right place for now.

dlaehnemann: So I'll stick it right here into sample.rs, then.

src/call/pairwise.rs
johanneskoester: Hi, sorry for the delay, we had two deadlines yesterday. I would have time today. If you like, please choose a time before 3pm.

dlaehnemann:
So, I have finally come around to addressing the remaining requests you had and rebasing the whole thing onto the current master.

I already had the CigarError changes implemented locally and have made any additions  to your implementation I had at their new home in rust-htslib's `src/bam/record.rs`.

With the refactoring that has happened in master, I had to adjust some parts substantially, so there are code parts where you might want syntax changes?

Also, I found a little bug in the `CigarString` parsing you use for overlap determination: It currently assumes the absence of HardClips in reads. It's marked with `Bug` and a `TODO` with a suggestion on how to handle it.

src/model/mod.rs
```diff
@@ -453,8 +507,8 @@ mod tests {
     #[test]
     fn test_same_pileup() {
         let variant = Variant::Deletion(3);
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.0)..AlleleFreq(1.0);
+        let tumor_all = ContinuousAlleleFreqs::inclusive( 0.0..1.0 );
+        let tumor_alt = ContinuousAlleleFreqs::inclusive( 0.0..1.0 ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: Yes.

src/model/mod.rs
```diff
@@ -665,9 +719,9 @@ mod tests {
             prior_model
         );
 
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.05)..AlleleFreq(1.0);
-        let tumor_ref = AlleleFreq(0.0)..AlleleFreq(0.001);
+        let tumor_all = ContinuousAlleleFreqs::inclusive( 0.0..1.0 );
+        let tumor_alt = ContinuousAlleleFreqs::inclusive( 0.05..1.0 ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: No.

src/model/mod.rs
```diff
@@ -665,9 +719,9 @@ mod tests {
...
+        let tumor_ref = ContinuousAlleleFreqs::inclusive( 0.0..0.001 ); // TODO: should this be (0.0)..(0.0) instead?
```
johanneskoester: No.

src/model/mod.rs
```diff
@@ -811,8 +865,8 @@ mod tests {
     fn test_example2() {
         let (case_obs, control_obs, model) = setup_example("tests/example2.obs.txt", 0.01, 0.03);
 
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.05)..AlleleFreq(1.0);
+        let tumor_all = ContinuousAlleleFreqs::inclusive( 0.0..1.0 );
+        let tumor_alt = ContinuousAlleleFreqs::inclusive( 0.05..1.0 ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: No.

src/model/mod.rs
```diff
@@ -833,8 +887,8 @@ mod tests {
     fn test_example3() {
         let (case_obs, control_obs, model) = setup_example("tests/example3.obs.txt", 0.01, 0.03);
 
-        let tumor_all = AlleleFreq(0.0)..AlleleFreq(1.0);
-        let tumor_alt = AlleleFreq(0.05)..AlleleFreq(1.0);
+        let tumor_all = ContinuousAlleleFreqs::inclusive( 0.0..1.0 );
+        let tumor_alt = ContinuousAlleleFreqs::inclusive( 0.05..1.0 ); // TODO: should this be left_exclusive() instead?
```
johanneskoester: No

src/model/priors/single_cell_bulk.rs
```diff
@@ -107,15 +113,67 @@ impl SingleCellBulkModel {
...
+            self.n_bulk_min
+        } else if *n >= self.n_bulk_max {
+            self.n_bulk_max
+        } else {
+            *n
+        }
+    }
+
+    /// determine values of k to use with current n and given allele frequency range
+    fn determine_k_b(&self, af: &ContinuousAlleleFreqs, n: &usize) -> (usize,usize) {
```
johanneskoester: Why the ref on usize? It is a copy-type, you can simply use `usize`.

src/model/sample.rs
```diff
@@ -324,11 +323,21 @@ impl Sample {
                     let end_pos = cigar.end_pos() as u32;
 
                     if pos <= start && end_pos >= start {
-                        let cigar = record.cigar();
-                        observations.push(
-                            self.read_observation(&record, &cigar, start, variant, chrom_seq)?
-                        );
-                        n_overlap += 1;
+                        if let Some( (prob_ref, prob_alt) ) = evidence::reads::prob_snv(record, &cigar, start, variant, chrom_seq)? {
```
johanneskoester: Any reason to not put this into read_observations again? It shares a lot of the code...

src/model/sample.rs
```diff
@@ -340,7 +349,8 @@ impl Sample {
                     let pos = record.pos() as u32;
                     let end_pos = cigar.end_pos() as u32;
 
-                    let overlap = {
+                    let overlap = { // BUG: SoftClips can be preceded by HardClips at read starts, followed by HardClips at read ends!
+                        // TODO: suggestion for removing the above BUG: consider this in the CigarStringView::end_pos calculation by adding in the include_softclips argument there, as well, and create an according function CigarStringView::start_pos
```
johanneskoester: Actually this is intended. I am only interested in the part of the read where the sequence is known and real. Hard clipped bases are not in the bam. So, I think the comments can be removed.

tests/lib.rs
```diff
@@ -130,17 +130,17 @@ fn call_tumor_normal(test: &str, exclusive_end: bool) {
     let events = [
         libprosic::call::pairwise::PairEvent {
             name: "germline".to_owned(),
-            af_case: AlleleFreq(0.0)..AlleleFreq(1.0),
+            af_case: ContinuousAlleleFreqs::inclusive( 0.0..1.0 ), // TODO: double-check, if this is meant to be inclusive
```
johanneskoester: Yes, these can all be left-exclusive.

src/model/sample.rs
dlaehnemann: I initially did it that way, but figured that it complicates read_observations() quite a bit, having to pass on the possible `None` from prob_snv() and therefor artificially wrapping the observation from an indel in an `Option<>`, as well. So I think it is the cleaner solution this way, but if you are e.g. expecting similar reasons for skipping reads in indels, as we have to for SNVs when the variant position is deleted ('D') or ref skipped ('N'), I can quickly merge this back together to facilitate such future changes.

src/model/sample.rs
dlaehnemann: Ok, I guess my explanation wasn't clear: The problem is the comparison to the first and last cigar operation, as SoftClips can at the beginning of the read be preceded by and at the end of the read followed by HardClips. I.e. if you have a CigarString that starts with `5H2S...`, the comparison `Cigar::SoftClip(l) = cigar[0]` will return `false` and you'll miss making the desired `pos`-adjustment for the `SoftClip` at the second Cigar position. Am I making sense? Or am I missing something that is going on in this code?

I guess, this could actually be avoided by one simple extra check here:
```
let pos = if let Cigar::SoftClip(l) = cigar[0] |
    ( Cigar::HardClip(_) = cigar[0] & Cigar::SoftClip(l) = cigar[1] ) {
```

dlaehnemann: (and of course a similar change for `end_pos`)

johanneskoester: Ah, now I get it, sorry! Yes, that makes absolute sense. Can you modify the statements like you propose in this PR?

src/model/sample.rs
johanneskoester: I think always returning `Some(obs)` in the indel case is fine.

src/model/sample.rs
johanneskoester: Wait, I will do that myself now. I need this at a different place as well. You can simply remove your comments.

dlaehnemann: Right on time, was just typing it out in the code... ;)

dlaehnemann:
Here come the latest requested changes, `read_observation()` now is back to doing the job for both SNVs and indels. Also, I pushed two previous commits, that I seem to have forgotten to push last week, and the BUG notice is gone, as you said you're on it right now. Any more points?

src/model/priors/single_cell_bulk.rs
```diff
@@ -107,15 +113,67 @@ impl SingleCellBulkModel {
...
+
+    fn prior_single(&self, _: AlleleFreq, _: &Variant) -> LogProb {
+        // TODO: stick in meaningful prior for the single cell sample, e.g. P(theta_b > 0 | bulk_pileup)
+        LogProb::ln_one()
+    }
+
+    /// Function to cap the use of the single cell amplification bias model at a coverage of 100,
+    /// as the Lodato et al. model was fit with coverages capped at 60 and starts behaving
+    /// weirdly above 100
+    fn cap_n_s(&self, n: &usize) -> usize {
```
johanneskoester: Here is still a reference to usize.

johanneskoester: Apart from that, perfect!

dlaehnemann:
Thanks for catching that one!

johanneskoester:
Great work, thanks!

dlaehnemann:
Well, thanks for your patience and the thorough code review! Unless I find any bugs, I'll mostly be on ProSolo for the coming days and probably will only come back to adding libprosic functionality for single cells in a little while. A libprosic hackathon for generalising core functionality not withstanding!

PR #7, opened 2017-05-09
Fix FDR calculation.

johanneskoester:


johanneskoester:
Sorry for breaking prosolo with this, but I figured it is important to do this refactoring of fdr control as soon as possible. This solution is much cleaner to use.

PR #9, opened 2017-05-24
Pair HMM based indel likelihood

johanneskoester:
Further, this PR changes the handling of `Variant` from copy to reference based. @dlaehnemann sorry that I had to do such an extensive change will your PR is not yet merged. However, I couldn't avoid this anymore because I needed to include  the insertion sequence into the corresponding variant type. I hope the conflicts are manageable. Also, I modified the `prob_read_snv` function to use a new helper from rust-htslib. The helper basically uses your implementation from PR #6. Hence, the results should be the same.


dlaehnemann:
No worries, that's basically my bad, not having been able to get to the changes that we discussed. I should be able to get a little of that done now, on Friday and then next week I'll be spending a lot of time on libprosic and prosolo. Thanks for always keeping me posted on the important changes.

PR #12, opened 2017-09-22
Refactoring, finetuned pair HMM implementation, removed artifacts from weird mapper decisions

johanneskoester:


johanneskoester:
Seems there are some network problems on travis... I'll restart this later.

Issue #16, opened 2017-11-23
separate libprosic-based CLI tool for controlling the FDR 


dlaehnemann:
In order to avoid duplicating too much functionality across `prosic2` and `prosolo` (and potential future `libprosic`-based variant callers), we should create a separate `libprosic`-based CLI tool to control the FDR for any `libprosic`-based caller, i.e. based on any `Event` spaces defined in accordance with `libprosic`.

dlaehnemann:
With the new varlociraptor setup, this is obsolete.

PR #17, opened 2017-11-23
add utils.rs function to filter by FDR thresholds

dlaehnemann:
Trying to filter the BCF file based on compound false discovery rate (FDR) thresholds for `Event` sets that were generated by the `control-fdr` subcommand of prosolo, I realised that `bcftools` cannot easily filter on (useful) combinations of multiple PHRED scores. So I implemented this functionality as the backend of an `apply-fdr` subcommand for prosolo.

In the midterm, it might be useful to generate a separate CLI tool only for controlling the false discovery rate, because this should be agnostic of the libprosic CLI tool used for generating the `Event` probabilities. I created issue #16 to track this idea.

Also, could we make this a release 0.5.1, as it effectively closes a gap in the FDR functionality from version 0.5.0? Or would semantic versioning rather require a release 0.6.0?

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -238,7 +238,7 @@ pub fn filter_by_threshold<E: Event>(
         }
 
         let variants = (utils::collect_variants(&mut record, false, false, None, false))?;
-        let mut events_probs = Vec::with_capacity(7);
+        let mut events_probs = Vec::with_capacity( variants.len() * tags.len() );
         for tag in &tags {
             if let Some(event_probs) = (record.info(tag.as_bytes()).float())? {
                 //tag present
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -238,7 +238,7 @@ pub fn filter_by_threshold<E: Event>(
         }
 
         let variants = (utils::collect_variants(&mut record, false, false, None, false))?;
-        let mut events_prob_sum = LogProb::ln_zero();
+        let mut events_probs = Vec::new();
         for tag in &tags {
             if let Some(event_probs) = (record.info(tag.as_bytes()).float())? {
                 //tag present
@@ -247,11 +247,12 @@ pub fn filter_by_threshold<E: Event>(
                         if !variant.is_type(vartype) || event_prob.is_nan() {
                             continue;
                         }
-                        events_prob_sum = events_prob_sum.ln_add_exp( LogProb::from( PHREDProb( *event_prob as f64 ) ) );
+                        events_probs.push( LogProb::from( PHREDProb( *event_prob as f64 ) ) );
                     }
                 }
             }
         }
+        let events_prob_sum = LogProb::ln_sum_exp(&events_probs);
         if events_prob_sum >= lp_threshold { out.write(&record)? };
     }
 }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -238,7 +238,7 @@ pub fn filter_by_threshold<E: Event>(
         }
 
         let variants = (utils::collect_variants(&mut record, false, false, None, false))?;
-        let mut events_probs = Vec::new();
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -218,6 +218,44 @@ pub fn collect_prob_dist<E: Event>(
...
+        let mut events_prob_sum = LogProb::ln_zero();
+        for tag in &tags {
+            if let Some(event_probs) = (record.info(tag.as_bytes()).float())? {
+                //tag present
+                for (variant, event_prob) in (&variants).into_iter().zip(event_probs.into_iter()) {
+                    if let Some(ref variant) = *variant {
+                        if !variant.is_type(vartype) || event_prob.is_nan() {
+                            continue;
+                        }
+                        events_prob_sum = events_prob_sum.ln_add_exp( LogProb::from( PHREDProb( *event_prob as f64 ) ) );
```
johanneskoester: Numerically it should be more stable to collect all probs into a vec and then apply `ln_sum_exp`.

dlaehnemann: Done, thanks for the hint. This might also solve a problem I saw on a bigger dataset, where the assertion of `p>=0.0` failed -- I am hoping this was a numerical issue and will double check if this now goes away.

src/utils.rs
```diff
@@ -218,6 +218,45 @@ pub fn collect_prob_dist<E: Event>(
...
+        if let Err(e) = calls.read(&mut record) {
+            if e.is_eof() {
+                return Ok(());
+            } else {
+                return Err(Box::new(e));
+            }
+        }
+
+        let variants = (utils::collect_variants(&mut record, false, false, None, false))?;
+        let mut events_probs = Vec::new();
```
johanneskoester: Sorry to be pedantic, but could you use `Vec::with_capacity` here?

dlaehnemann: Sure, please be. Just learned something new about Rust. :)

src/utils.rs
```diff
@@ -218,6 +218,45 @@ pub fn collect_prob_dist<E: Event>(
...
+        if let Err(e) = calls.read(&mut record) {
+            if e.is_eof() {
+                return Ok(());
+            } else {
+                return Err(Box::new(e));
+            }
+        }
+
+        let variants = (utils::collect_variants(&mut record, false, false, None, false))?;
+        let mut events_probs = Vec::with_capacity(7);
```
johanneskoester: Mhm, the 7 looks suspicious to me. Why 7? Shouldn't it be `variants.len() * tags.len()`?

dlaehnemann: You're right, I was only thinking of the current maximum of events we have (7 in ProSolo), but didn't think of possible multiple variants per site and of just quickly calculating this once per site. Done now.

src/utils.rs
```diff
@@ -218,6 +218,45 @@ pub fn collect_prob_dist<E: Event>(
     Ok(prob_dist)
 }
 
+pub fn filter_by_threshold<E: Event>(
```
johanneskoester: Ok, now we need a good doc comment above. I promise, this is the last request!

dlaehnemann: Don't promise something you might not be able to keep... ;)

I appreciate the code review very much, so please do request further changes if you see something!

johanneskoester:
Thanks :-)!

johanneskoester:
We can do a new release, but maybe you can also put `libprosic = { git = "https://github.com/prosic/libprosic.git" }` into your `Cargo.toml`.

dlaehnemann:
Yes, I'll do that for testing for now, just in case I find further bugs. But eventually I'd also like to push out a new release of prosolo to bioconda, to keep analyses as reproducible as possible. I'll let you know.

Thanks!

PR #20, opened 2017-12-07
abstract Event probabilities sum into function

dlaehnemann:
* use numerically more stable `ln_sum_exp()` on vector with collection of Event probs

**Please don't merge, until the following is addressed:**

A second goal (further commits needed!) is to resolve numerical issues in summing up `Event` probabilities. @johanneskoester, would you have some recommendations regarding the following issue and my idea to resolve it?

I have now factored out the summing up of `Event`probabilities into a separate function that also is used in control-fdr and that first collects the probabilities at a site into a vector and the runs `ln_sum_exp()` on it. Nevertheless, control-fdr chokes on some sites, where the sum of the probabilities is slightly above of certain `Event`s is above one, e.g. the following line:

```
chr1	62976472	rs1168043	T	G	5682.93	PASS	AN=2;AC=2;PROB_HOM_REF=177.487;PROB_ADO_TO_REF=30.2547;PROB_ADO_TO_ALT=148.039;PROB_HOM_ALT=0.807449;PROB_ERR_ALT=2094.59;PROB_HET=12.1714;PROB_ERR_REF=9.66315;CASE_AF=1;CONTROL_AF=1	GT	1/1
```

If I sum up all `Event`s indicating the presence of an alternative allele (`ADO_TO_REF + ADO_TO_ALT + HOM_ALT + HET + ERR_REF`) with the following expression in `gnome-calculator` (which I assume suffers from similar numerical issues):
```
10^(-30,2547/10) + 10^(-148,039/10) + 10^(-0,807449/10) + 10^(-12,1714/10) + 10^(-9,66315/10)
```

I get out: `1,00000045890689`

And this fails in the following line of `libprosic` code:
https://github.com/PROSIC/libprosic/blob/1664d33cd43fd476275a06a913a38285f960f30a/src/estimation/fdr/ev.rs#L47

Which eventually triggers this panic (where the `p` passed down into this is in natural log space and thus a probability <= 1 should be `p <= 0`):
https://github.com/rust-bio/rust-bio/blob/76726e94a2cde8fb7a227ac60a4cb9bc6a9f62e8/src/stats/probs/mod.rs#L31

I think, we'll have to include an `epsilon` somewhere, to allow for such slight overshoots in probability sums, and I currently have three alternative ideas where to implement this -- the first two would be in `rust-bio`, the last right here in `libprosic`:

1. Ensure, that a `LogProb` is always `<= 0` upon construction of any `LogProb`, offering the standard existing constructor(s), but adding an assert there and then also adding alternative constructors that allow a certain `epsilon` and cut the probability into the correct range if it is outside but within the `epsilon` range.
2. Create an alternative `ln_one_minus_exp()` where we implement the `epsilon` that allows to move such slight overshoots onto the limits of the probability range.
3. Implement such an `epsilon` check and cut in `libprosic`, e.g. in `collect_prob_dist()` right here:
https://github.com/PROSIC/libprosic/blob/1664d33cd43fd476275a06a913a38285f960f30a/src/utils.rs#L231-L232

Is this reasonably clear? Any thoughts?

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -201,7 +201,7 @@ fn tags_prob_sum(
             }
         }
     }
-    Ok(LogProb::ln_sum_exp(&tags_probs_out))
+    Ok( LogProb::ln_sum_exp(&tags_probs_out).cap_numerical_overshoot(0.000001) )
 }
 
 /// Collect distribution of posterior probabilities from a VCF file that has been written by
@@ -276,10 +276,37 @@ mod tests {
     use super::*;
 
     use rust_htslib::bcf;
-    use bio::stats::Prob;
+    use bio::stats::{Prob, LogProb};
     use model::VariantType;
     use ComplementEvent;
     use SimpleEvent;
+
+    #[test]
+    fn test_tags_prob_sum() {
+
+        // set up test input
+        let test_file = "tests/resources/test_tags_prob_sum/overshoot.vcf";
+        let overshoot_calls = bcf::Reader::from_path( test_file ).unwrap();
+        let mut record = bcf::Record::new();
+        overshoot_calls.read(&mut record);
+
+        // set up all alt events with names as in prosolo
+        let alt_tags = [
+            String::from("PROB_ADO_TO_REF"),
+            String::from("PROB_ADO_TO_ALT"),
+            String::from("PROB_HOM_ALT"),
+            String::from("PROB_HET"),
+            String::from("PROB_ERR_REF")
+        ];
+
+        let snv = VariantType::SNV;
+
+        if let Ok( prob_sum ) = tags_prob_sum(&mut record, &alt_tags, &snv) {
+            assert_eq!( LogProb::ln_one(), prob_sum );
+        } else {
+            panic!("tags_prob_sum(&overshoot_calls, &alt_events, &snv) returned Error")
+        }
+    }
 
     #[test]
     fn test_collect_prob_dist() {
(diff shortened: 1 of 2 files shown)


johanneskoester:
Yes, I know this proplem, and it is what you can expect from numerical uncertainties. Indeed, a function in rust-bio is a good idea. What about 
```
impl LogProb {
    pub fn fix(&self, epsilon: f64) -> LogProb { ... }
}
```

This way, you can also fix an existing LogProb. I am unsure about the right name though. fix is certainly a bit to generic.

dlaehnemann:
OK, I'll create a branch and a pull request over at `rust-bio` with a reference to this pull request, and then we can discuss details of exact name and scope of the function over there.

dlaehnemann:
Any objections to the current state of this one? Otherwise I would very much welcome a merge and a new libprosic release, to also be able to create a new prosolo release and bundle it up on bioconda.

johanneskoester:
I will try to do the release today once I am back at my laptop.

dlaehnemann:
Thanks already!

And I could also do a release, if that's easier. My suggestion for the change log would be:

* added functionality to filter by FDR thresholds based on sets of `Events` and improved code for summing up Event likelihoods (incl. fixing numerical overshoot issues)
* dependency updates (`rust-htslib` to `0.16`, `rust-bio` to `0.16` from `rustc-serialize` to `csv` and `serde`)
* some test fixes

johanneskoester:
Yes, that would be great, thanks!

PR #21, opened 2018-01-03
move from rustc-serialize dep to csv 1.0.0-beta.5

dlaehnemann:
This mirrors an update in `rust-bio 0.15` compared to `rust-bio 0.14` (I will need this update and actually an update to the next future release of `rust-bio` to enable the new `cap_numerical_overshoot()` function for `LogProb`s from `rust-bio` here in `libprosic`).

@johanneskoester:
While this branch compiles properly, 4 tests fail with a FloatParsingError on field 8 of the first record in the observation csv files example[2347].obs.txt in the new csv function `deserialize()` (<https://docs.rs/csv/1.0.0-beta.5/csv/struct.Reader.html#method.deserialize>).
Any ideas what might have changed in comparison to the `decode()` function in csv 0.15.0 (<https://docs.rs/csv/0.15.0/csv/struct.Reader.html#method.decode>)? It sounds like something to do with the `collect()` into the `Observation` `struct`, especially the `Evidence` `enum` in it, but the above cited documentation doesn't indicate that any of this shouldn't work, so I'm giving up for now -- the only alternative I would currently see would be to follow the examples for the new `deserialize()` function that would treat each row/record separately and manually add each to one of two collections `control_obs` or `case_obs`, but that doesn't seem elegant...

### DIFF
--- a/src/call/pairwise.rs
+++ b/src/call/pairwise.rs
@@ -14,6 +14,7 @@ use bio::io::fasta;
 use model::AlleleFreqs;
 use model::priors;
 use model::PairCaller;
+use model::sample::Evidence;
 use model;
 use Event;
 use utils;
@@ -77,6 +78,18 @@ fn pileups<'a, A, B, P>(
     Ok(pileups)
 }
 
+#[derive(Debug, Serialize)]
+struct ObservationRecord {
+    chrom: String,
+    pos: u32,
+    allele: u32,
+    sample: String,
+    prob_mapping: LogProb,
+    prob_alt: LogProb,
+    prob_ref: LogProb,
+    prob_mismapped: LogProb,
+    evidence: Evidence
+}
 
 /// Call variants with the given model.
 ///
@@ -145,11 +158,6 @@ pub fn call<A, B, P, M, R, W, X, F>(
 
     let mut outobs = if let Some(f) = outobs {
         let mut writer = try!(csv::WriterBuilder::new().delimiter(b'\t').from_path(f) );
-        // write header for observations
-        writer.write_record(
-            ["chrom", "pos", "allele", "sample", "prob_mapping",
-             "prob_alt", "prob_ref", "prob_mismapped", "evidence"].iter()
-         )?;
         Some(writer)
     } else { None };
     let mut record = bcf::Record::new();
@@ -179,11 +187,34 @@ pub fn call<A, B, P, M, R, W, X, F>(
                 for (i, pileup) in pileups.iter().enumerate() {
                     if let &Some(ref pileup) = pileup {
                         for obs in pileup.case_observations() {
-                            outobs.serialize((chrom, record.pos(), i, "case", obs))?;
+                            outobs.serialize(
+                                ObservationRecord {
+                                    chrom: chrom.to_string(),
+                                    pos: record.pos(),
+                                    allele: i as u32,
+                                    sample: "case".to_string(),
+                                    prob_mapping: obs.prob_mapping,
+                                    prob_alt: obs.prob_alt,
+                                    prob_ref: obs.prob_ref,
+                                    prob_mismapped: obs.prob_mismapped,
+                                    evidence: obs.evidence.clone()
+                                }
+                            )?;
                         }
                         for obs in pileup.control_observations() {
-                            outobs.serialize((chrom, record.pos(), i, "control", obs))?;
-                        }
+                            outobs.serialize(
+                                ObservationRecord {
+                                    chrom: chrom.to_string(),
+                                    pos: record.pos(),
+                                    allele: i as u32,
+                                    sample: "control".to_string(),
+                                    prob_mapping: obs.prob_mapping,
+                                    prob_alt: obs.prob_alt,
+                                    prob_ref: obs.prob_ref,
+                                    prob_mismapped: obs.prob_mismapped,
+                                    evidence: obs.evidence.clone()
+                                }
+                            )?;}
                     }
                 }
                 outobs.flush()?;
--- a/src/call/pairwise.rs
+++ b/src/call/pairwise.rs
@@ -144,9 +144,9 @@ pub fn call<A, B, P, M, R, W, X, F>(
     };
 
     let mut outobs = if let Some(f) = outobs {
-        let mut writer = try!(csv::Writer::from_file(f)).delimiter(b'\t');
+        let mut writer = try!(csv::WriterBuilder::new().delimiter(b'\t').from_path(f) );
         // write header for observations
-        writer.write(
+        writer.write_record(
             ["chrom", "pos", "allele", "sample", "prob_mapping",
              "prob_alt", "prob_ref", "prob_mismapped", "evidence"].iter()
          )?;
@@ -179,10 +179,10 @@ pub fn call<A, B, P, M, R, W, X, F>(
                 for (i, pileup) in pileups.iter().enumerate() {
                     if let &Some(ref pileup) = pileup {
                         for obs in pileup.case_observations() {
-                            outobs.encode((chrom, record.pos(), i, "case", obs))?;
+                            outobs.serialize((chrom, record.pos(), i, "case", obs))?;
                         }
                         for obs in pileup.control_observations() {
-                            outobs.encode((chrom, record.pos(), i, "control", obs))?;
+                            outobs.serialize((chrom, record.pos(), i, "control", obs))?;
                         }
                     }
                 }
--- a/src/call/pairwise.rs
+++ b/src/call/pairwise.rs
@@ -78,19 +78,6 @@ fn pileups<'a, A, B, P>(
     Ok(pileups)
 }
 
-#[derive(Debug, Serialize)]
-struct ObservationRecord {
-    chrom: String,
-    pos: u32,
-    allele: u32,
-    sample: String,
-    prob_mapping: LogProb,
-    prob_alt: LogProb,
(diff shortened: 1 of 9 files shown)


johanneskoester:
I have no idea either... Maybe write a separate test case that just tries to deserialize an observation. Also, have a look how the observation would be serialized now. I guess this is where the difference comes in.

dlaehnemann:
After another day of digging into the documentation of the `csv` crate and the new `serialize` and `deserialize`, I think I have properly updated the dependency and the respective code. And all tests are now passing. But please do review, especially for style (I e.g. introduced two `clone()`s in `pairwise.rs`, which didn't feel right, but worked).

In effect, the new version of csv doesn't like nested `structs` and I eventually got it to work by explicitly doing the `serialize()` and `deserialize()` on flat `structs` -- well, mostly flat, with the `Evidence` `enum` as the last entry. Also, when specifying a `struct` for the record structure, headers are printed automatically, and upon reading matched automatically.

dlaehnemann:
Sorry, @johanneskoester, if I keep on bugging you with this--it's a prerequisite for two other things I would like to get in asap (one of which is getting Pull Request #20 merged). If there is anything else I can do to facilitate a review, please do let me know.

src/call/pairwise.rs
```diff
@@ -179,11 +187,34 @@ pub fn call<A, B, P, M, R, W, X, F>(
                 for (i, pileup) in pileups.iter().enumerate() {
                     if let &Some(ref pileup) = pileup {
                         for obs in pileup.case_observations() {
-                            outobs.encode((chrom, record.pos(), i, "case", obs))?;
+                            outobs.serialize(
```
johanneskoester: So, it now cannot deal with nested structs in tuples anymore? This sounds like a serious downside to me. Is it not possible to stick with the old csv version, while the rust-bio crate is using the new one?

johanneskoester: According to BurntSushi/rust-csv#75 nested stuff should work in theory, just not with automatic header writing? Or is that an old version before the switch to serde?

src/model/sample.rs
```diff
@@ -160,7 +160,7 @@ impl RecordBuffer {
 
 
 /// An observation for or against a variant.
-#[derive(Clone, Debug, RustcDecodable, RustcEncodable)]
```
dlaehnemann: From what I understand, we cannot have a dependency on  a different version of `csv` in our setup, as the compiler complains about `rust-bio` >= `0.15` as:

```
error[E0277]: the trait bound `bio::stats::LogProb: rustc_serialize::Decodable` is not satisfied
   --> src/model/sample.rs:163:24
    |
163 | #[derive(Clone, Debug, RustcDecodable, RustcEncodable)]
    |                        ^^^^^^^^^^^^^^ the trait `rustc_serialize::Decodable` is not implemented for `bio::stats::LogProb`
    |
    = note: required by `rustc_serialize::Decodable::decode`
```

As I understand it, the use of the `LogProb`s in here, which are using the newer `csv` version in `rust-bio` to do their deriving (with `serde`), we cannot use the older `csv` version for deriving Traits, here. Or am I missing something?

src/call/pairwise.rs
dlaehnemann: As for writing nested structs: From what I gather from the [new `serialize` docs here](https://docs.rs/csv/1.0.0-beta.5/csv/struct.Writer.html#structs), `struct` entries have to be named (which ours are) and all scalars (which our `enum` isn't). So unless I am missing something, we cannot use automatic header generation with our setup, unless we use the redefined `struct` I suggested.

But as in the link you provided, we could just go back to using `has_header(false)` in the `WriterBuilder` and explicitly writing the header line. Then, it can handle the nesting for writing. I have just tested this and could quickly clean it up and check it in, right away. Would you prefer that?

dlaehnemann:
Thanks for looking at the code and commenting. Is the above suggestion for the `csv` writing viable? And are you OK with the current `csv` reading setup in the tests of `sample.rs` and `model/mod.rs`?

src/model/sample.rs
johanneskoester: Ok, makes sense.

src/call/pairwise.rs
johanneskoester: Yes, this is a good solution. Just write the header manually, and then we are able to use the previous way of writing the observation.

src/model/sample.rs
```diff
@@ -608,13 +608,43 @@ mod tests {
...
     fn read_observations(path: &str) -> Vec<Observation> {
-        let mut reader = csv::Reader::from_file(path).expect("error reading example").delimiter(b'\t');
-        let obs = reader.decode().collect::<Result<Vec<(String, u32, u32, String, Observation)>, _>>().unwrap();
-        let groups = obs.into_iter().group_by(|&(_, _, _, ref sample, _)| {
-            sample == "case"
-        });
-        let case_obs = groups.into_iter().next().unwrap().1.into_iter().map(|(_, _, _, _, obs)| obs).collect_vec();
+        let mut reader = csv::ReaderBuilder::new().delimiter(b'\t').from_path(path).expect("error reading example");
+        let mut case_obs = vec![];
+        for row in reader.deserialize() {
```
johanneskoester: I would prefer to still deserialize into `(String, u32, u32, String, Observation)`. Is that possible?

dlaehnemann: When trying to unpack as similar to the previous code as possible, i.e. with:
```
let unpacked = reader.deserialize().collect::<Result<Vec<(String, u32, u32, String, Observation)>, _>>().unwrap();
```

It panics with:
```
'called `Result::unwrap()` on an `Err` value: Error(Deserialize { pos: Some(Position { byte: 79, line: 2, record: 1 }), err: DeserializeError { field: Some(8), kind: ParseFloat(ParseFloatError { kind: Invalid }) } })', /checkout/src/libcore/result.rs:859
```

The documentation indicates, that automatic reading into structs will only work when `has_headers(false)` is set for the `ReaderBuilder`. However, no other method to skip the first line in reading is provided, so we can only do avoid choking on the first line on the iterator level, e.g. like this:

```
let mut reader = csv::ReaderBuilder::new().has_headers(false).delimiter(b'\t').from_path(path).expect("error reading example");
let deserialized = reader.deserialize().skip(1);
let unwrapped = deserialized.into_iter().collect::<Result<Vec<(String, u32, u32, String, Observation)>, _>>().unwrap();
```

But this still chokes on the unwrap, with a different error:
```
panicked at 'called `Result::unwrap()` on an `Err` value: Error(Deserialize { pos: Some(Position { byte: 79, line: 2, record: 1 }), err: DeserializeError { field: None, kind: Message("invalid type: unit variant, expected newtype variant") } })'
```

I'm assuming, this has to do with the `enum Evidence` not being tagged, because this Error goes away when I annotate the `enum` with:
```
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(untagged)]
pub enum Evidence {
    /// Insert size of fragment
    InsertSize(String),
    /// Alignment of a single read
    Alignment(String)
}
```

However, this then chokes on the first line with a number for an`InsertSize` in it instead of the String `Alignment` with:
```
panicked at 'called `Result::unwrap()` on an `Err` value: Error(Deserialize { pos: Some(Position { byte: 3748, line: 51, record: 50 }), err: DeserializeError { field: None, kind: Message("data did not match any variant of untagged enum Evidence") } })'
```

And I'm not sure where to go with that... No clue how to give it a hint about the type based on what it reads from the file, there. Any suggestions?

johanneskoester: Ok, I see. This is crap. I suggest we do the following.
1. We keep the workaround with `Row`. It is not so much of an issue, because it is only in the test cases.
2. If you find the time, you could maybe create a minimal example and post an issue in the csv github repo, such that the author becomes aware of it.

dlaehnemann: Yep, I'll try to create that minimal example tomorrow in the morning. Tuple or struct containing a struct that contains such an ambiguous enum and the corresponding csv file to go with it.

dlaehnemann: Just for cross-reference, issues raised upstream:
https://github.com/BurntSushi/rust-csv/issues/100
https://github.com/BurntSushi/rust-csv/issues/101

Feel free to chime in or correct me over there! ;)

johanneskoester: Thanks! All fine!

dlaehnemann: 1. It looks like automatic deserializing of `csv` input previously written out from `csv`-serialized version of our `enum Evidence` will not be supported (see discussion in <https://github.com/BurntSushi/rust-csv/issues/100>). With the current functionality, we could get to a `csv`-supported version by making the variant arguments distinct types, e.g. making it:
```
pub enum Evidence {
    InsertSize(f64),
    Alignment(String)
}
```

2. Once this is sorted, we could get around the header matching that does not work in the Tuple `.collect()` encompassing both primitives and the `struct Observation` by using `.has_headers(false)` in the `csv::ReaderBuilder` and a `.skip(1)` after the `.deserialize()` (see discussion in <https://github.com/BurntSushi/rust-csv/issues/101>).

With 1. we avoid the current use of the helper `fn recode_evidence(string: String) -> Evidence` and with 2. we would get around defining the helper `struct Row`. But this would involve changing `enum Evidence`, and I'm not sure if that would be desirable--although I don't really see, why it currently is `InsertSize(String)`, when the variant argument should usually be a number? Should we make this change, or leave the code as is?

PR #22, opened 2018-01-12
New Version

dlaehnemann:
Sorry, no write access for me on the master branch--so I'll go through this branch and see how far I get.

PR #23, opened 2018-01-12
quick patch (omit double header for FDR thresholds)

dlaehnemann:
I found a little glitch from my downstream tests in prosolo: The new csv serializer automatically outputs the header automatically, based on field names. With the manual header output in our case, we got a double header and this patch restores previous behaviour. Should merge this before PR #22, which should be rebased after this is merged.


dlaehnemann:
@johanneskoester Sorry, but I have to ask for a quick approval of the merge of this branch. It only patches an unwanted and superfluous header line that snuck into the `csv` output of fdr thresholds.

Issue #18, opened 2017-11-23
calculate and emit `Event` likelihoods for sites without candidate alt allele


dlaehnemann:
To get an unbiased evaluation of libprosic-based calling of single nucleotide variant `Event`s, these should not only be called where the candidates provide a possible alternative allele, but also at sites that are listed in the candidates file, but where no alternative allele is listed.

From what I see, this would require explicitly parsing those sites in the `utils.rs` function `collect_variants()` with some useful base/nucleotide provided to the SNV type (ambiguity code for any other nucleotide than the reference one at that site?) and in `model/evidence/reads.rs` the function `prob_read_base()` would then need to handle that base/nucleotide.

Does this make sense?

johanneskoester:
I wam not yet sure that I understand why that would be needed. Could you explain why?

dlaehnemann:
Sure, my current use case is:

I have a `candidate.bcf`file that contains single nucleotide variant / genotype calls. In addition to the common calls where at least one alternative allele is present, I also have a lot of sites where no candidate alternative allele is present in the `candidate.bcf`, e.g.:
```
chr1    65446   .       C       <*>     0       .       DP=223;AC=0;AN=2       GT      0/0
```

As I am not only interested in genotypes at sites with an alternative allele present, but also in genotypes at sites where no alternative allele was present in the set used for candidate generation--in my case a gold standard dataset separate from the actual samples genotyped. So I would like to just run the likelihood model on these sites, as well, to get a benchmarking of the genotyping capabilities that is not biased against homozygous reference sites. Does this make more sense?

And as we have no alternative allele at those sites, we need to create one when we parse these sites from the BCF. To me, a dedicated nucleotide Type based on the [IUPAC ambiguity code](http://www.bioinformatics.org/sms/iupac.html) seems like the cleanest solution--we would just need to create the respective equality operation on this type in `rust-bio`, probably here:
https://github.com/rust-bio/rust-bio/blob/dba57123d158e23cc74de3335efab9fbc0309f1a/src/alphabets/dna.rs

I think this should work without any great overhead in parsing and computation?!

As an alternative, we could just generate a random non-reference genotype that we then calculate the likelihoods against, but we might loose read information / `Evidence` if we have different deviating reads at a site -- and considering all alternative nucleotides bundled up should give a conservative likelihood for the homozygous reference genotype, which sounds preferable to me.

johanneskoester:
[...]

PR #24, opened 2018-02-05
Calls for hom ref sites

dlaehnemann:
This branch introduces a version of what we are looking for in issue #18. It introduces a new variant type `Variant::Ref` that captures no other info than that the site is suspected to be (homozygous) reference. In effect, we only test for the presence of any alternative allele (i.e. nucleotide) instead of testing for a specific allele (which we would, if the candidate file suggested some evidence for this).

It contains a test for the new `prob_ref()` function, but still lacks a test on a proper VCF file with `<*>` entries.


dlaehnemann:
This branch now contains the above mentioned test, so if Travis agrees, I'd love to see this merged soon.

Also, together with PR #26 I would go for another release, to be able to pull the new functionality into ProSolo.

johanneskoester:
I have some doubts about breaking the symmetry here. Can we have a call today, to discuss this in person? Also, note that you can use this branch from prosolo via Cargo.toml. No need to hurry with merge and release.

src/model/mod.rs
```diff
@@ -89,15 +89,17 @@ impl Deref for ContinuousAlleleFreqs {
 pub enum VariantType {
     Insertion(Option<Range<u32>>),
     Deletion(Option<Range<u32>>),
-    SNV
+    SNV,
+    Ref, // homozygous reference site
```
johanneskoester: Can we call this `None` instead?

dlaehnemann: Jep, will refactor and double-check `pairwise.rs` later today. Will also add `TODO`s for looking at this again for (no-)indel calling -- or would you prefer this filed as an issue instead of in the code?

johanneskoester: I don't think we can do much in this regard. But yes, a TODO sounds reasonable.

dlaehnemann:
I have refactored now, and hand over to Travis.

Regarding the BCF parsing, from what I found, the parsing of alternative alleles is done in `pairwise.rs` via the function `collect_variants()` within `pileups()`:
https://github.com/PROSIC/libprosic/blob/1e285348e6e87cf825042e9960799380753c0d88/src/call/pairwise.rs#L65

This function is defined in `src/utils.rs`, where the parsing of the additional alternative allele `<*>` is added. Does this address your `BCF` parsing question from above?

Issue #25, opened 2018-02-05
arbitrary ALPHAS via command-line or finer resolution of small FDRs?


dlaehnemann:
I have the case where I would like to further resolve FDRs smaller than 0.01, and the current definition only goes down to that value:

https://github.com/PROSIC/libprosic/blob/e3da0d58729a29bfcb3a246ecacf6986f78ff7ff/src/estimation/fdr/mod.rs#L7-L36

As a first step, I would suggest adding the following values to this static definition of ALPHAS (unless there is some good reason that `0.01` is the smallest value, e.g. a methodological limitation?!):
```
0.00001,
0.00005,
0.0001,
0.0005,
0.001,
0.005,
```

In a second step (or maybe directly?), we could allow for passing extra (or all) alpha values on the command line.

Any objections? Any further considerations to be made?

johanneskoester:
Makes sense, feel free to create a PR for the extra hardcoded values. Making this configurable by allowing `--alphas 0.01 0.05 ...` is a good idea as well, but we can indeed do that later.

dlaehnemann:
I just realised that Rust only auto-implements the Trait `IntoIterator` for arrays of length up to 32. This currently leaves me only 4 extra values that I can add. Would it be OK to remove current entries `0.65` and `0.75` in order to add more resolution closer to zero? In my data, the are around these higher values does not see any change of FDR filter level any more...

dlaehnemann:
My suggestion in PR #26 now is to remove two more higher FDR points in exchange for more resolution in the very low range.

dlaehnemann:
P.S.: I'd keep this issue open beyond PR #26 to remind us of the command-line option plan.

dlaehnemann:
This is basically resolved with PR #45

PR #26, opened 2018-02-06
add FDR resolution closer to zero

dlaehnemann:
* trades off some resolution closer to 1, where not much should happen
  with filtering thresholds, anyways (beyond a certain point, all calls
  are included, anyways) -- the trading off is necessary, as Rust
  currently only implements the `IntoIterator` trait for arrays up to
  length 32
* adjusts the respective expected output in the `fdr_ev_1` test


dlaehnemann:
This looks at issue #25.

dlaehnemann:
If you agree with this approach, feel free to merge the PR once the tests pass.

dlaehnemann:
Thanks!

PR #35, opened 2018-06-20
Likelihood cache

dlaehnemann:
Hey @johanneskoester,

I'm stuck here, so I pushed my most promising current attempt. This is just trying to get the structure including the mutable likelihood cache to work in the `tumor_normal` model, so I currently commented out the `flat` and the `single-cell-bulk` model.

Having `PairPileup` in the signature of `PairModel` functions seems slightly circular to me, but the main problem I currently have is the compiler error I get about PairPileup not being `Sized`. Maybe you have a quick idea or insight to help me here?

thanks and cheers,
david

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -497,12 +497,6 @@ impl<'a, A: AlleleFreqs, B: AlleleFreqs, P: priors::PairModel<A, B>> PairPileup<
         if self.marginal_prob.get().is_none() {
             debug!("Calculating marginal probability.");
 
-            let case_likelihood = |af_case: AlleleFreq, af_control: Option<AlleleFreq>| {
-                self.case_likelihood(af_case, af_control)
-            };
-            let control_likelihood = |af_control: AlleleFreq, af_case: Option<AlleleFreq>| {
-                self.control_likelihood(af_control, af_case)
-            };
             let p = self.prior_model.marginal_prob(self);
             debug!("Marginal probability: {}.", p.exp());
 
@@ -513,13 +507,6 @@ impl<'a, A: AlleleFreqs, B: AlleleFreqs, P: priors::PairModel<A, B>> PairPileup<
     }
 
     pub fn joint_prob(&self, af_case: &A, af_control: &B) -> LogProb {
-        let case_likelihood = |af_case: AlleleFreq, af_control: Option<AlleleFreq>| {
-            self.case_likelihood(af_case, af_control)
-        };
-        let control_likelihood = |af_control: AlleleFreq, af_case: Option<AlleleFreq>| {
-            self.control_likelihood(af_control, af_case)
-        };
-
         let p = self.prior_model.joint_prob(&self, af_case, af_control);
         p
     }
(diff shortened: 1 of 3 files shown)


src/model/mod.rs
```diff
@@ -458,6 +459,9 @@ pub struct PairPileup<'a, A, B, P> where
     prior_model: &'a P,
     case_sample_model: likelihood::LatentVariableModel,
     control_sample_model: likelihood::LatentVariableModel,
+    // these two caches are useful for PairModels where case and control are independent of each other and Events are just combinations of across the two axes, as e.g. in the single-cell-bulk model
+    case_likelihood_cache: &'a mut BTreeMap<AlleleFreq, LogProb>,
+    control_likelihood_cache: &'a mut BTreeMap<AlleleFreq, LogProb>,
```
johanneskoester: These have to be owned by PairPileup. I.e., don't use a reference here, just `BTreeMap<AlleleFreq, LogProb>`. Have a look at this again: https://doc.rust-lang.org/book/second-edition/ch04-01-what-is-ownership.html#ownership-rules
PairPileup must own the caches.

johanneskoester: I think the sized issue might go away once this is fixed.

src/model/mod.rs
```diff
@@ -460,8 +460,8 @@ pub struct PairPileup<'a, A, B, P> where
     case_sample_model: likelihood::LatentVariableModel,
     control_sample_model: likelihood::LatentVariableModel,
     // these two caches are useful for PairModels where case and control are independent of each other and Events are just combinations of across the two axes, as e.g. in the single-cell-bulk model
-    case_likelihood_cache: &'a mut BTreeMap<AlleleFreq, LogProb>,
-    control_likelihood_cache: &'a mut BTreeMap<AlleleFreq, LogProb>,
+    case_likelihood_cache: Cell<BTreeMap<AlleleFreq, LogProb>>,
+    control_likelihood_cache: Cell<BTreeMap<AlleleFreq, LogProb>>,
```
johanneskoester: Why the cell?

dlaehnemann: Oh, someone is watching... ;)

The idea was to keep the PairPileup in general immutable, and only make the fields mutable that need to be. A side product of something I tried when just removing the referencing here wasn't allowed by the compiler as valid syntax (i.e. neither of the following compiled:
```
    case_likelihood_cache: 'a mut BTreeMap<AlleleFreq, LogProb>,
    case_likelihood_cache: mut BTreeMap<AlleleFreq, LogProb>,
```
And while it could definitely be that I just don't get the syntax right, the `Cell` solution nevertheless seems to make more sense to me.

dlaehnemann:
This code gets me a bit further, but eventually I once again run into:
```
error[E0525]: expected a closure that implements the `Fn` trait, but this closure only implements `FnMut`
   --> src/model/priors/tumor_normal.rs:148:27
    |
148 |             let density = |af_tumor| {
    |                           ^^^^^^^^^^ this closure implements `FnMut`, not `Fn`
...
151 |                 pileup.case_likelihood(af_tumor, Some(af_normal))
    |                 ------ closure is `FnMut` because it mutates the variable `pileup` here
...
157 |                 LogProb::ln_simpsons_integrate_exp(&density, *af_tumor.start, *af_tumor.end, self.grid_points)
    |                 ---------------------------------- the requirement to implement `Fn` derives from here

error: aborting due to previous error

For more information about this error, try `rustc --explain E0525`.
error: Could not compile `libprosic`.
```

But I'm still on it...

johanneskoester:
The newest rust-bio should solve that for you. For that, you also need to remove the `&` before `density` in the call to `ln_simpsons_integrate_exp`.

src/model/mod.rs
johanneskoester: Just this: `case_likelihood_cache: BTreeMap<AlleleFreq, LogProb>,`. And then, `&mut self` in methods that change the cache. As a rule of thumb: interior mutability (Cell) should only be used if it cannot be avoided, because you (a) generate runtime overhead, and (b) constrain potential parallelism we might add in the future.

dlaehnemann:
This PR turned out to be a dead end and is retired in favour of PR #36.

PR #36, opened 2018-06-25
Likelihood caches via mutable pair pileup passing

dlaehnemann:
This introduces two likelihood caches into `PairPileup`, to avoid calculating the same likelihoods multiple times. So far this is only implemented for the `TumorNormalModel`, to first make it work before applying it to the other models, as well.

@johanneskoester: I hope this is a cleaner restart compared to before. I have tried around a lot of things and this is closest to working. The compiler now chokes on the two following spots (each with a line ahead shown for context, relevant is the bottom line):

https://github.com/PROSIC/libprosic/blob/6cb3d6d0d6be1d17476cad5e53dfef6d8726f81e/src/call/pairwise.rs#L203-L204

https://github.com/PROSIC/libprosic/blob/6cb3d6d0d6be1d17476cad5e53dfef6d8726f81e/src/call/pairwise.rs#L239-L240

In the current code state, it throws:
```
error[E0596]: cannot borrow immutable borrowed content `*pileup` as mutable
   --> src/call/pairwise.rs:241:49
    |
240 |                 if let &Some(ref pileup) = pileup {
    |                              ---------- consider changing this to `ref mut pileup`
241 |                     let (case_af, control_af) = pileup.map_allele_freqs();
    |                                                 ^^^^^^ cannot borrow as mutable
```

If I follow the recommendation of adding `mut`, it instead complains:
```
error[E0596]: cannot borrow anonymous field `(pileup as serde::export::Some).0` of immutable binding as mutable
   --> src/call/pairwise.rs:240:30
    |
240 |                 if let &Some(ref mut pileup) = pileup {
    |                              ^^^^^^^^^^^^^^ cannot mutably borrow field of immutable binding

```

So as far as I understand it, it is some kind of serialization issue that makes `pileup` an anonymous field. But l can't get my head around what to do here -- make the binding mutable? the field non-anonymous?

Any suggestions?

### DIFF
--- a/src/call/pairwise.rs
+++ b/src/call/pairwise.rs
@@ -175,7 +175,7 @@ pub fn call<A, B, P, M, R, W, X, F>(
 
         // translate to header of the writer
         outbcf.translate(&mut record);
-        let pileups = pileups(
+        let mut pileups = pileups(
             &inbcf, &mut record, pair_model, &mut reference_buffer,
             omit_snvs, omit_indels, max_indel_len, exclusive_end
         )?;
@@ -200,8 +200,8 @@ pub fn call<A, B, P, M, R, W, X, F>(
             // write posterior probabilities
             let mut posterior_probs = Array::default((events.len(), pileups.len()));
             for (i, event) in events.iter().enumerate() {
-                for (j, pileup) in pileups.iter().enumerate() {
-                    let p = if let &Some(ref pileup) = pileup {
+                for (j, pileup) in pileups.iter_mut().enumerate() {
+                    let p = if let &mut Some(ref mut pileup) = pileup {
                         // use joint probability instead of posterior since we do the
                         // normalization below, turning joint probabilities into posteriors.
                         Some(pileup.joint_prob(&event.af_case, &event.af_control))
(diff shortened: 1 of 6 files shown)


dlaehnemann:
Another idea would be to make the entries in the `pileups` vector mutable. As far as I can see, this would require making the part inside the Some() on line 71 mutable:
https://github.com/PROSIC/libprosic/blob/6cb3d6d0d6be1d17476cad5e53dfef6d8726f81e/src/call/pairwise.rs#L68-L77

Maybe this implies having to make the enclosing `pileups()` function a method of `PairPileup`? Or should the constructor of `PairPileup` return something like a `&mut Self`? Sorry if this seems slightly confused, I just don't seem to get anywhere with thinking about how to get the mutability syntax right...

johanneskoester:
You need to iterate over mutable references of pileups. It is impossible to convert an immutable to a mutable reference. So, in the fist case, iter_mut(), in the second case &mut pileups.

dlaehnemann:
Thanks @johanneskoester!

I had tried the `iter_mut()` and all kinds of mutability syntax, but apparently didn't get it right and didn't find the right spots from the resulting compiler errors. With these two recommendations and following the compiler errors I have now gotten all the spots. Will now extend this to the other `PairModel`s.

dlaehnemann:
@johanneskoester: This is now updated to include all the implemented `PairModel`s with the new cache and the respective mutability of `PairPileup` and its `_likelihood` functions, and passing only the `pileup` itself around, not individual fields of it.

The code compiles and passes all active tests both locally and on `Travis CI`. But please do have a critical scroll through the code changes. Some spots include explicit scoping with `{ }` to clearly separate mutable borrows of different `PairPileup` fields. And if you agree with the changes, feel free to merge.

johanneskoester:
Once you have a feeling for the combined speedups of this and our other changes, please let me know.

PR #38, opened 2018-07-23
Use binary MAPQ if requested

johanneskoester:
It turns out that MAPQ is biased towards lower MAPQ for non-ref fragments when observing SVs. Hence, when considering MAPQ in out latent variable model, we always get a bias towards the ref allele, which generates both false negatives and false positives (either missing a variant in tumor or in normal).

To overcome this problem, we can binarize the MAPQ, taking the maximum observed (e.g. 60) for all reads with MAPQ > 0 and 0 otherwise. This works really well, giving superior precision-recall over all other callers.


tests/lib.rs
```diff
@@ -658,6 +643,7 @@ fn test_sc_bulk_indel() {
     call_single_cell_bulk("test_sc_bulk_indel", true, "chr1");
     let mut call = load_call("test_sc_bulk_indel");
     check_info_float(&mut call, b"CONTROL_AF", 0.0, 0.0);
+    // TODO case AF seems to be > 0, also from looking at IGV. Are you sure the test case is correct?
```
johanneskoester: @dlaehnemann this test case fails here, although I cannot see any piece of code affecting it... Moreover, are you should the assertion of CASE_AF=0 was correct before? When looking at IGV, it looks like >0 to me...

dlaehnemann: Funny one...

You are right, that there are reads supporting the indel in `single_cell.bam` for that test case. The maximum likelihood allele frequency could nevertheless still be `0.0` and must have been in the past. But if I remember correctly, I just went for some random indel example from existing test files and set the expected output to the one I got for this example, so we could change the expected output if we feel confident about it.

Though, one little thing that I noticed in reading the respective testing code, is that there is a minor error introduced in the test case setup with the change in allele frequency syntax:
https://github.com/PROSIC/libprosic/blame/abf168eff435c21b4fa604848629a36e61489eb3/tests/lib.rs#L266

This should instead read like the `Event`s setup in ProSolo:
https://github.com/ProSolo/prosolo/blob/f42ccfe9343b3cf5837851533ea42f9feb48bce4/src/call.rs#L136

This would affect the total sum of the calculated likelihoods and thus the marginalisation of them -- and might in turn tip over the maximum likelihood allele frequency?!

dlaehnemann: Done, let's see if this passes.

johanneskoester: Same AF. So, if you feel like the estimate is right, I would vote for updating the test case.

dlaehnemann: Ah, it's the `CONTROL_AF` that's failing (and that has been failing before) as non-zero with around 0.12. And without any amplification bias in the bulk sample, this estimate makes perfect sense, with 3 out of 23 reads carrying the indel allele. So I agree with updating the test case -- please feel free, otherwise I'll pick this up tomorrow.

I'm just wondering why the test passed before...?!

And also slightly, why I picked that position and this combination of `bam`-files as a test case. But I guess, I was just going for proper basic candidate indel parsing and handling in the single cell bulk model.

dlaehnemann: I have now adjusted both CONTROL_AF and PROB_HET, and have also included PROB_HOM_REF in the test. From looking at the alignments, the new values do make sense and seem trustworthy. 

I am just wondering how the previous values came about and why the tests passed. But then, I haven't used the indel functionality for single cell data so far.

PR #40, opened 2018-07-25
Likelihood caches working

dlaehnemann:
Just as a quick explanation for the verbose new code: I first tried just substituting `or_insert()` with `or_insert_with()`, but the `Rust` ownership model doesn't allow that. `self` needs to be borrowed mutably in that context to allow the insertion to take place, and the call to `self.case_sample_model.likelihood_pileup()` is another (immutable) borrow while the other borrow is still active. Thus the funny scoping of `l` and all...

Also, this code leads to an ~80% runtime reduction in ProSolo:
https://github.com/ProSolo/prosolo/issues/2#issuecomment-407439622

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -530,14 +530,13 @@ impl<'a, A: AlleleFreqs, B: AlleleFreqs, P: priors::PairModel<A, B>> PairPileup<
     ) -> LogProb {
         // no af_control given, because case and control are independent
         if af_control.is_none() {
-            let l: LogProb;
+            // get likelihood if already cached
             if let Some(likelihood) = self.case_likelihood_cache.get(&af_case) {
-                // get likelihood if already cached
-                return *likelihood
-            } else {
-                // compute otherwise
-                l = self.case_sample_model.likelihood_pileup(&self.case, af_case, None);
+                return *likelihood;
             }
+            // compute otherwise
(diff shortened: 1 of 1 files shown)


src/model/mod.rs
```diff
@@ -530,10 +530,16 @@ impl<'a, A: AlleleFreqs, B: AlleleFreqs, P: priors::PairModel<A, B>> PairPileup<
...
-            *self.case_likelihood_cache.entry(af_case)
-                // compute and insert otherwise
-                .or_insert( self.case_sample_model.likelihood_pileup(&self.case, af_case, None) )
+                return *likelihood
+            } else {
+                // compute otherwise
+                l = self.case_sample_model.likelihood_pileup(&self.case, af_case, None);
+            }
+            self.case_likelihood_cache.insert(af_case, l);
+            l
```
johanneskoester: Does this not work?
```
if let Some(likelihood) = self.case_likelihood_cache.get(&af_case)  {
    return likelihood;
}
let l = self.case_sample_model.likelihood_pileup(&self.case, af_case, None);
self.case_likelihood_cache.insert(af_case, l);
l
```

dlaehnemann: It does, and it is shorter. So tests are running!

dlaehnemann:
Feel free to merge and then delete the branch. Next up will be some flamegraphing utility.

PR #41, opened 2018-07-27
Alignment properties

johanneskoester:
1. Automatically estimate global alignment properties over the first 10000 reads of the BAM file (TODO: choose a random subsample?).
2. Use these for insert size, and for alt sampling probability.
3. Cleanup now unused code related to `observations::Common`.


johanneskoester:
@dlaehnemann these changes provide another row of significant improvements of indel calling quality. They will unfortunately also affect how prosolo calls libprosic. You have to create an `AlignmentProperties` object first. See the single cell part of `tests/lib.rs` for an example how to do that.

PR #43, opened 2018-08-16
Flamegraph annotations

dlaehnemann:
This provides flamegraphing utilities for libprosic on two possible levels, that can be requested as `--features` during compilation of the library (or of a binary that uses `libprosic`):

* `flame_it`: some higher level functions are annotated with the respective `#[cfg_attr(feature = "flame_it", flame)]`feature annotation (enabled by the [flamer crate](https://github.com/llogiq/flamer)) , meaning if `--features flame_it` is requested at compile time, every call of the function will be recorded with start and end (enabled by the [flame crate](https://github.com/TyOverby/flame))
* `flame_it_details` annotates some functions deeper down and is an annotation that could also be used in library dependencies of `libprosic`, e.g. through the respective pull requests in [`rust-bio`](https://github.com/rust-bio/rust-bio/pull/187) and [`rust-htslib`](https://github.com/rust-bio/rust-htslib/pull/105)

For an example how to implement this into binary using libprosic, have a look at [a respecitve issue](https://github.com/ProSolo/prosolo/issues/2) and the commits leading up to [this one](https://github.com/ProSolo/prosolo/commit/37df2e19e3dcd0d03eacf2e89e76f4da2c3f9480) over at [ProSolo](https://github.com/ProSolo/prosolo).

I expect that this will only pass tests, once the above mentioned PRs at `rust-bio` and `rust-htslib` are merged. Also, this points to the latest master branch commits on those projects, which should be changed to some fixed released version before the next release...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,7 @@ approx = "0.1"
 rusty-machine="0.4"
 ordered-float = "0.5"
 ndarray = "0.6"
-# flame = { version = "0.2.2", optional = true }
-flame = "0.2.1-pre2"
+flame = { version = "0.2.1-pre2", optional = true }
 flamer = { version = "^0.2.3", optional = true }
 quick-error = "1.2"
 vec_map = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,9 @@ approx = "0.1"
 rusty-machine="0.4"
 ordered-float = "0.5"
 ndarray = "0.6"
-flame = { version = "^0.1.2", optional = true }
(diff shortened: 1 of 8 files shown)


dlaehnemann:
Flamegraphing internally with the currently available rust crates is not as flexible as we would like. Instead I recently got `perf` + `stackcollapse-perf.pl` + `flamegraph.pl` to work with some extra tricks. So I retire this branch in favor of this solution:
<https://gist.github.com/dlaehnemann/df31787c41bd50c0fe223df07cf6eb89>

PR #47, opened 2018-10-18
ContinuousAlleleFreqs of points should be inclusive or exclusive

dlaehnemann:
at both ends, one-sided exclusivity does not make sense.


johanneskoester:
Thanks!

PR #51, opened 2018-11-12
Precompute pairhmm cell probs

johanneskoester:
This PR introduces precomputed miscall probabilities into pairHMM computation. By that, we should save maybe 20 cycles in the innermost loop of the HMM. @dlaehnemann since you have the setup already, can you please create a flamegraph of this vs master branch? You can use test09 in lib.rs for that.


johanneskoester:
Thank you so much!

johanneskoester:
Closing in favor of #53, which contains the changes from this branch.

PR #53, opened 2018-11-16
Use fastexp branch in rust-bio

johanneskoester:
@dlaehnemann may I ask you to create another flamegraph of this branch? Again for test09?

Thanks a lot!

PR #54, opened 2018-11-21
PairHMM optimizations

johanneskoester:
@dlaehnemann this is another round of optimizations in Rust-Bio. Again this PR includes the changes from #51 and #53. May I ask you for another flamegraph?

Thanks!

# Summary of overall optimizations
* Omit lower left and upper right triangle in pairHMM dynamic programming matrix if gap extension probability is zero. (#54)
* Try to skip as many LogProb operations as possible in pairHMM code. (#54)
* Use fast approximation of exp (#53).
* Use precomputed per-base match, mismatch and indel probabilities (#51).
* Calculate best edit distance with Myer's algorithm in linear time. Use that (times 2) as a bound for a banded pair-HMM. In particular for small indels (which usually cause performance bottlenecks because they are abundant), this should cause a huge speedup, because large parts of the matrix will be omitted (#54).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "GPL-3.0"
 include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "README.md", "LICENSE"]
 
 [dependencies]
-bio = "0.23"
+bio = { git = "https://github.com/rust-bio/rust-bio.git", branch = "fast-exp" }
 log = "0.3"
 rust-htslib = "0.22"
 GSL = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "GPL-3.0"
 include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "README.md", "LICENSE"]
 
 [dependencies]
-bio = { git = "https://github.com/rust-bio/rust-bio.git", branch = "fast-exp" }
+bio = { git = "https://github.com/rust-bio/rust-bio.git", branch = "optimize-pairhmm" }
 log = "0.3"
 rust-htslib = "0.22"
 GSL = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "GPL-3.0"
 include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "README.md", "LICENSE"]
 
 [dependencies]
-bio = { git = "https://github.com/rust-bio/rust-bio.git", branch = "optimize-pairhmm" }
+bio = { git = "https://github.com/rust-bio/rust-bio.git", branch = "banded-pairhmm" }
 log = "0.3"
 rust-htslib = "0.22"
 GSL = "1.0"
@@ -17,12 +17,11 @@ itertools = "0.6"
 itertools-num = "0.1"
 approx = "0.1"
 rusty-machine="0.4"
-ordered-float = "0.5"
+ordered-float = "1.0"
 ndarray = "0.6"
 quick-error = "1.2"
 vec_map = "0.8"
 regex = "0.2"
-
 serde = "1"
 serde_derive = "1"
 csv = "1.0.0-beta.5"
--- a/src/estimation/alignment_properties.rs
+++ b/src/estimation/alignment_properties.rs
@@ -8,7 +8,7 @@ use std::u32;
 use bio::stats::PHREDProb;
 use csv;
 use itertools::Itertools;
-use ordered_float::NotNaN;
+use ordered_float::NotNan;
 use rust_htslib::bam::{self, record::Cigar};
 use statrs::statistics::{OrderStatistics, Statistics};
 
@@ -40,7 +40,7 @@ impl AlignmentProperties {
     pub fn estimate<R: bam::Read>(bam: &mut R) -> Result<Self, Box<Error>> {
         let mut record = bam::Record::new();
         let mut tlens = Vec::new();
-        let mut max_softclip = NotNaN::new(0.0).unwrap();
+        let mut max_softclip = NotNan::new(0.0).unwrap();
         let mut max_del_cigar_len = 0;
         let mut max_ins_cigar_len = 0;
         let mut max_mapq = 0;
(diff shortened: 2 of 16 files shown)


johanneskoester:
@dlaehnemann hmm, there seems to be some kind of randomness in FDR correction. This test is not always failing: https://travis-ci.org/PROSIC/libprosic/builds/457961846#L691

dlaehnemann:
Hmm, weird one. It's off by one from the expected number of entries, so I guess that it's a corner case that flips. Maybe this is due to some minor numerical fluctuations that are host-system dependent? I can't get the test to fail on my machine. So I have just restarted the build on Travis CI to see whether it fails again. If it does, it would seem a bit more systematic and will have something to do with the host environment -- if it doesn't I am at a loss.

But in general, we could just relax the test a bit or try to find and remove the corner case entry for now. It's mostly a regression test, making sure that we don't break fdr control.

dlaehnemann:
So, still failing on Travis and not failing locally on my machine -- even with the newest rust and cargo versions just installed. So seems to be system-consistent. Any further ideas or thoughts?

dlaehnemann:
Also, here's the newest flamegraph:

https://gist.github.com/dlaehnemann/8de85fcf5f3801c822f87b267b53f19d

(to view the interactive and searchable SVG, right-click the preview in the gist and select View Image)

It seems to have drastically changed...

johanneskoester:
The changes are a bit unexpected... I would have thought that the ratio of prob_related rather goes down than up. Was that on a different machine? Can you also provide a timing?

dlaehnemann:
It's all done on the same machine, but I double-checked everything again and found a mistake: I had forgotten to put the following in the `Cargo.toml`:

```
# Uncomment the following if you need debugging symbols despite compiling for optimised performance
# with --release flag
[profile.release]
debug = true
```

With a rerun including this, the flamegraph has a reduced proportion of time spent in `prob_related`:

<https://gist.github.com/dlaehnemann/433f440a3c6afcd97a688eac0e2c695d>

And the timing of three consecutive runs is:

```
user	0m0.628s
user	0m0.620s
user	0m0.624s
```

So a further drastic reduction compared to the last timing (which is documented in #53 with an average of about `0,84s`).

To ease future flamegraphing, I would like to include the above code `[profile.release]` code with comment and all in the Cargo.toml permanently -- it should be commented out completely as the default and only activated for actual flamegraphing. But this way, I don't always have to copy-paste it in every time. Also, either just include it as a commit here, or I can create a separate branch for it, if you prefer not to have it come in as a side-effect of this branch...?!

johanneskoester:
Yes, good idea!

johanneskoester:
No need for a separate branch. Moreover, would a custom profile called flamegraph be possible?

johanneskoester:
Do you know why the stack over ln_sum_exp is suddenly not showing all those other calls from before anymore?

dlaehnemann:
> No need for a separate branch. Moreover, would a custom profile called flamegraph be possible?

Just had a quick look at the cargo docs and I would opt for permanently adding the `debug=true` config to the `[profile.bench]`:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections

It has all the same options set as the `[profile.release]`, but is also the profile applied when using `cargo test --release`. Which by the way is what I have been using for the profiling with `test09` for the changes in this PR, so I'm not sure if `debug=true` is even needed, because it had only been set for `[profile.release]`.

> Do you know why the stack over ln_sum_exp is suddenly not showing all those other calls from before anymore?

Hmm, I just realised, that I have been doing the latest flamegraph on a different machine -- confused myself, there. Will do it again on the same (desktop) machine, now.

dlaehnemann:
Once again, an updated flamegraph -- now really produced on the same desktop machine (to improve the documentation for me, I used the binary `target/release/deps/lib-bd13d8378779c7b3`):

<https://gist.github.com/dlaehnemann/250cabbf940cee7087ce7e4d7ee4e497>
(to view the interactive and searchable SVG, right-click the preview in the gist and select View Image)

And here goes the `time`ing on that machine, which is even better:

```
user	0m0,567s
user	0m0,548s
user	0m0,553s
```

dlaehnemann:
Also, for checking out the different profile configurations, here are the flamegraphs for:

1. no `debug=true` entry anywhere, using `cargo test --release` for compilation (`target/release/deps/lib-bcd8a0116f7c55d5 test09`):
<https://gist.github.com/dlaehnemann/91a8fef6bf0ddbc0889fe0d9fad79acd>
2. `debug=true` in `[profile.bench]`, using `cargo test --release` for compilation (`target/release/deps/lib-cff13dd7514feea8`):
<https://gist.github.com/dlaehnemann/58547eeba0d011fc12ab468bdbcf70a1>

Both of these variants lack some levels of granularity compared to the previous solution of running `cargo test --release` while including in `Cargo.toml`:
```
[profile.release]
debug = true
```

Also, I tried using `cargo bench` and `cargo test --benches` for compilation using the `[profile.debug]`, but can't then run `test09`, so I'm giving up on this one. But the cargo docs are definitely wrong on `cargo test --release` -- it uses the `[profile.release]`, not the `[profile.bench]`...

As custom profiles also haven't arrived, yet, I think we are back to using `[profile.release]` and commenting in and out, for now...

johanneskoester:
Yes, I agree. Regarding the test case: yes, let's just relax it a bit.

johanneskoester:
@dlaehnemann can we get another flamegraph and timing?

dlaehnemann:
> can we get another flamegraph and timing?

Of course we can, your royal highness... ;)

Binary (used on desktop machine) is:
```
target/release/deps/lib-0ce23af7a06e69d4
```
Produced from commit:
```
b7b3dc0b9e35bc527562632fc8cc9c07bb182671
```

Timing (looks like a further improvement):
```
user	0m0,479s
user	0m0,485s
user	0m0,457s
```

Flamegraph (looks like there are changes with `ln_sum_exp()`, which I think was the intention?):
<https://gist.github.com/dlaehnemann/f3e3e3c293b670d9530003d44166fded>
(to view the interactive and searchable SVG, right-click the preview in the gist and select View Image)

johanneskoester:
That's it. In a real scenario improvements should be even larger, in particular in case of many small indels.

johanneskoester:
May I ask you to fix the test case? I don't know if I was too tired or stupid, but I could not find the failing assertion...

johanneskoester:
Afterwards, we can merge and release a new version.

dlaehnemann:
The respective tests are still older versions than are in the current master HEAD (see here: <https://github.com/PROSIC/libprosic/blob/5097a5e5d35445cfb7723b41b9f5129fa10049c7/tests/lib.rs#L611-L627> and <https://github.com/PROSIC/libprosic/blob/5097a5e5d35445cfb7723b41b9f5129fa10049c7/tests/lib.rs#L332-L339>), so maybe this branch needs a rebase onto the current master?

johanneskoester:
Ok, done, thank you! I'll take care of the rest.

PR #56, opened 2018-12-11
cap_numerical_overshoot: increase acceptable overshoot to 0.0001

dlaehnemann:
After a quick discussion, @johanneskoester and I decided that we feel comfortable with capping a numerical overshoot in log-space of up to 1e-04. So that's the new hard-coded value -- exposing this as a command-line parameter to the filtering functionality in prosic and prosolo would require significant extra work.


johanneskoester:
0.7.3 is released.

PR #57, opened 2019-02-11
Various improvements

johanneskoester:
* Allow subclonal somatic mutations in the normal sample.
* Normalize allele probabilities in order to avoid bias by proximal variants. After normalization, probs are rescaled to the original magnitude by multiplying complements of the base qualities.
* Use max allele probability as a proxy for the probability of the case where the fragment is mismapped. Otherwise, the mismapping case dominates the likelihood if allele probabilities become small. The latter would cause a loss of sensitivity: we would miss low allele frequencies because the corresponding likelihoods are all considered as mismappers.
* Use full MAPQ again. The previous fix makes MAPQ binarization superfluous.
* Allow to use minimum observable allele frequencies in prior models. Further allow to set the minimum number of trusted observations (e.g. 2). By this, we get more robust event probabilties.
* Test cases have been relaxed. They now qualitatively check whether events are called correctly, and don't rely so much on particular probabilities.
* Do not consider insert size if the probability for one allele becomes zero and is not within one SD for the other allele. By this, we get rid of artifacts that are induced by weird insert sizes that dominate the probabilities of the other evidence types.


johanneskoester:
Finally!

PR #65, opened 2019-05-06
Grammar for arbitrary calling scenarios

johanneskoester:


johanneskoester:
@jafors this is the feature needed for Laura's project. You should be able to write the following yaml (something like that, don't remember the exact scenarios we discussed, they are on my desk):
```
samples:
  sample0:
    resolution: 100
  sample1:
    resolution: 100

events:
  de_novo:
    sample0: "0.0"
    sample1: "[0.1,1.0]"
  loss:
    sample0: "[0.1,1.0]"
    sample1: "0.0"
  absent:
    sample0: "0.0"
    sample1: "0.0"
  present:
    sample0: "]0.0,1.0]"
    sample1: "]0.0,1.0]"
```

and then run
`varlociraptor call variants generic scenario.yaml --bams sample0=path/to/bam sample1=path/to/other/bam ...

Issue #68, opened 2019-05-14
varlociraptor chokes on indels due to Myers pattern_matching


dlaehnemann:
From testing the circleci-provided bioconda package before merging the [varlociraptor recipe PR ](https://github.com/bioconda/bioconda-recipes/pull/15231) over there, I ran into this:

```
[src/model/modes/generic.rs:57] &posterior = GenericPosterior {
    resolutions: [
        100
    ]
}
[src/model/modes/generic.rs:58] &likelihood = GenericLikelihood {
    inner: [
        Normal(
            SampleLikelihoodModel
        )
    ]
}
thread 'main' panicked at 'Pattern too long', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bio-0.26.1/src/pattern_matching/myers/mod.rs:276:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:70
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:200
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:215
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:478
   5: std::panicking::begin_panic
   6: <bio::pattern_matching::myers::Myers<T>>::new
   7: <varlociraptor::model::evidence::reads::IndelEvidence as varlociraptor::model::evidence::reads::AbstractReadEvidence>::prob
   8: varlociraptor::model::sample::Sample::fragment_observation::{{closure}}
   9: varlociraptor::model::sample::Sample::extract_observations
  10: <varlociraptor::call::Caller<A, L, Pr, Po, ModelPayload>>::call
  11: varlociraptor::cli::run
  12: varlociraptor::main
  13: std::rt::lang_start::{{closure}}
  14: std::panicking::try::do_call
             at src/libstd/rt.rs:49
             at src/libstd/panicking.rs:297
  15: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:87
  16: std::rt::lang_start_internal
             at src/libstd/panicking.rs:276
             at src/libstd/panic.rs:388
             at src/libstd/rt.rs:48
  17: main
  18: __libc_start_main
[...]

PR #69, opened 2019-05-14
Constrain indel window size

dlaehnemann:
The constrain of the input parameter seems to be working, but there are still corner cases that fail -- e.g. the newly included test `pattern_too_long`. To keep up to date with the failing tests of newer commits, I would move discussion from issue #68 over here.


johanneskoester:
Thanks a lot for the excellent groundwork and the test case @dlaehnemann! This helped a lot to fix the errors.

dlaehnemann:
Well, thanks for the `testcase` options, that made this really easy.

PR #74, opened 2019-07-31
Use interpretable priors for transition between gains in CNV calling.

johanneskoester:


johanneskoester:
This is not the way to go. A follow-up PR with a new approach that is currently being worked on will come hopefully soon.

Issue #78, opened 2019-08-21
Testcase for spurious somatic variant call


FelixMoelder:
[testcase.zip](https://github.com/varlociraptor/varlociraptor/files/3525493/testcase.zip)

Issue #77, opened 2019-08-21
Adaptive smallest value for allele freq intervals regardless of resolution


johanneskoester:
If the resolution is too low (e.g. 5) we are unable to see events with too few supporting reads. Therefore, Varlociraptor should always add an allele frequency >0.0 that is derived from 1/n with n being the pileup size.

johanneskoester:
Fixed already.

PR #79, opened 2019-08-21
testcase for a spurious somatic tumor variant that should intuitively…

johanneskoester:
… be a somatic normal


johanneskoester:
xref #78

johanneskoester:
xref #77

PR #90, opened 2019-11-13
Implement contig-specific universe definition.

johanneskoester:
Also add corresponding test case.


johanneskoester:
cc @christopher-schroeder

PR #99, opened 2020-02-19
improve insert size estimation, esp. handling when no usable records found

dlaehnemann:
This PR does several things regarding the insert size estimation from the first 10000 (usable) records of a BAM file:

1. It more systematically skips unusable records to make sure it at least tries to look at 10000 fragments before calculating an insert size mean and sd.
2. Once 100.000 reads are skipped, it stops the estimation, because this indicates that something is wrong with the reads (or they are all single-ended).
3. If no reads up to that point were usable, it will explicitly set the InsertSize `mean` and `sd` to `std::f64::NAN`. This is now caught in the sample creation and sets `use_fragment_evidence` to `false` and avoids trying to use an insert size when that value is NAN. This was the original bug that started this investigation.

And another important change:
**This PR changes from calculating an `insert_size` median and storing it into `insert_size.mean` to actually calculating a mean.** @johanneskoester: if the mean calculation was intended, please feel free to revert this change (and maybe rename to `insert_size.median`;).


src/estimation/alignment_properties.rs
```diff
@@ -83,42 +85,104 @@ impl AlignmentProperties {
         let mut tlens = Vec::new();
         let mut max_mapq = 0;
         let mut i = 0;
+        let mut skipped = 0;
+        let mut n_soft_clip = 0;
+        let mut n_not_useable = 0;
         while i <= 10000 {
+            if skipped >= 100000 {
+                eprint!(
```
johanneskoester: Please use the `warn!` macro here.

src/estimation/alignment_properties.rs
```diff
@@ -83,42 +85,104 @@ impl AlignmentProperties {
...
-        let mut valid = tlens
-            .into_iter()
-            .filter(|l| *l <= upper && *l >= lower)
-            .collect_vec();
-        properties.insert_size.mean = valid.median();
-        properties.insert_size.sd = valid.iter().std_dev();
-
-        Ok(properties)
+        if tlens.len() == 0 {
+            eprint!(
```
johanneskoester: `warn!` macro.

src/estimation/alignment_properties.rs
```diff
@@ -83,42 +85,104 @@ impl AlignmentProperties {
...
+                       - marked as a duplicate\n\
+                       - mate mapped to different template (e.g. different chromosome)\n\
+                       - failed some quality check according to the 512 SAM flag\n\
+                       - mate unmapped\n\
+                       - record unmapped\n",
+                nu = n_not_useable,
+                sc = n_soft_clip,
+                nr = skipped
+            );
+            properties.insert_size.mean = std::f64::NAN;
```
johanneskoester: The idiomatic way is to use Option<f64> instead, and set it so `None` here. This enforces all code that uses the property object to specifically handle the None case, instead of accidentally calculating stuff using the NaN values.

dlaehnemann: Just to check that I understand correctly, before implementing this... I would wrap the `InsertSize` into an option and then make it a `None` here and a `Some(InsertSize)` everywhere else?

So for the general deifition:
```
pub struct AlignmentProperties {
    insert_size: Option<InsertSize>,
    pub(crate) max_del_cigar_len: u32,
    pub(crate) max_ins_cigar_len: u32,
    pub(crate) frac_max_softclip: f64,
}
```

And here:
```
properties.insert_size = None;
```

The rest of the property object may well be usable, so we would not want to set the whole `properties` to `None`, right?

johanneskoester: Yes, exactly.

dlaehnemann:
I have addressed your points as discussed above, so this should be good to go with another quick glance.

PR #105, opened 2020-03-22
New variant trait system and refactoring.

johanneskoester:
@dlaehnemann I plan to extend this for breakend support. After discussion with @christopher-schroeder, I think implementing breakends will allow us to reduce everything else to it (except SNV and MNV) (thanks a lot Christo!). It is however complicated as hell, and requires a major rewrite. I'll try to do this myself in the next one or two weeks.


johanneskoester:
@dlaehnemann @tedil the new variant trait system is now in place.

johanneskoester:
I just need to fix the final clippy warnings.

PR #106, opened 2020-03-25
Added linker arguments for macOS

FelixMoelder:
This PR adds a config file with linker arguments required for building on macOS.


FelixMoelder:
There seems to be an issue with the code coverage token which does not concern the PR itself.

PR #108, opened 2020-05-27
use two scales for tmb plot (for beyond and before VAF 0.1). Offer three different plotting modes.

johanneskoester:
Namely, histogram, curve, and stratified area.


johanneskoester:
cc @jafors

PR #114, opened 2020-07-24
LOH calling

dlaehnemann:
Using the coin-cbc ILP solver, this code should now do some usable LOH calling. I open this PR to check, whether the updated GitHub Action works (installing `coincbc` from conda-forge and activating it) and to enable a proper review of the code! Will further test this code on real data, now.


dlaehnemann:
@johanneskoester 
I have now tried it on real data with the `coincbc` solver on a single chromosome of medium size (8). It core dumps after using a lot of memory (more than 200 GB RAM when we control the FDR of germline heterozygosity and more than 300 GB RAM if we don't -- but I'm not sure if peak memory was maybe higher and might have caused the termination) and running for almost 2 hours. This does not seem promising.
Now I am trying the GLPK solver, but I don't have too much hope for this approach.

Maybe we have to think about a divide-and-conquer strategy? Or switch over to my more heuristic suggestion, which should achieve a similar maximization but with much less space and time requirements?

PR #124, opened 2020-09-10
remove possibility of using alignment_properties.insert_size when it is None

dlaehnemann:
When running varlociraptor on some single end read data, I got this panic!() that I had put there a while ago:
```
AlignmentProperties has insert_size set to `None`, but you are trying to use it without checking for None.This is a bug.
```

The backtrace indicates, that the problem is in `varlociraptor::variants::sample::SampleBuilder::alignments()`:
```
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:217
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:526
  11: std::panicking::begin_panic
  12: varlociraptor::variants::sample::SampleBuilder::alignments
  13: crossbeam_utils::thread::ScopedThreadBuilder::spawn::{{closure}}
```

But I just removed the `insert_size()` function and replaced it with a public instance of `insert_size` in `AlignmentProperties` and then try to handle every case where an `insert_size` is used. Please feel free to correct any of the behaviour I put there, be it the handling I implemented or the `panic!()`s that I put wherever I had no clue what else to do with a `None` right there.

And one point I still do not get is what the `--omit_insert_size` flag actually does. I don't think that it actually prevents the usage of the `insert_size` information...?!

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -505,10 +505,11 @@ pub fn run(opt: Varlociraptor) -> Result<()> {
 
                     // If we omit the insert size information for calculating the evidence, we can savely allow hardclips here.
                     let allow_hardclips = omit_insert_size;
-                    let alignment_properties = est_or_load_alignment_properites(
+                    let alignment_properties = est_or_load_alignment_properties(
                         &alignment_properties,
                         &bam,
-                        allow_hardclips,
+                        omit_insert_size,
+                        allow_hardclips
                     )?;
 
                     let gap_params = GapParams {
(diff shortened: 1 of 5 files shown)


johanneskoester (approved):
Sorry for being so late with this. LGTM.

PR #132, opened 2020-11-11
allow ignoring strand bias and read orientation bias

dlaehnemann:
This PR introduces the flags `--omit-strand-bias` and `--omit-read-orientation-bias` and handles them accordingly.

@johanneskoester Currently, this does not include a test for either of them or both of them being set. Should I simply create a test case for each based on the version in this branch?


dlaehnemann:
I added test cases for all the combinations of setting `--omit-read-orientation-bias` and `--omit-strand-bias`, including handling the respective options in `testcase.yaml` and defaulting to `false` for both flags. I added the options manually into the files, as this setup seemed like the smallest change to the testing harness.

Please note that I had to update to `GSL = "2.0.1"` for cargo to succeed compiling locally on my machine. But as tests pass nicely, this should all be fine.

src/variants/model/bias/mod.rs
```diff
@@ -74,27 +74,35 @@ impl BiasesBuilder {
 impl Biases {
     pub(crate) fn all_artifact_combinations(
         consider_read_orientation_bias: bool,
-    ) -> Box<dyn Iterator<Item = Self>> {
-        if consider_read_orientation_bias {
-            Box::new(
+        consider_strand_bias: bool,
+    ) -> Option<Box<dyn Iterator<Item = Self>>> {
```
johanneskoester, line 78: When using [`std::iter::empty()`](file:///home/johannes/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/std/iter/fn.empty.html) for the case where both args are false, you can avoid the Option type.

johanneskoester (changes requested):
Awesome! Just one thing.

src/variants/model/bias/mod.rs
dlaehnemann: Ah, cool. I was looking for something like this, but didn't know how and where to look. A classical Rust thing...

dlaehnemann: Done.

PR #148, opened 2021-04-12
TMB plots for multiple samples

jafors:
This PR seeks to enhance the TMB plotting functionality towards a multi-sample barplot. 
 - Vartypes have been reduced to Signatures (T>C falls under the same category as A>G etc.), but the original code is still present - maybe it can be an optional choice which representation to use.
- For visibility, TMB in the multi-sample plot is shown for a minimal VAF value which can be chosen as a parameter.

This is just a first functional version to get a quick representation of multiple samples side by side.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,7 +6,6 @@
 use std::collections::HashMap;
 use std::convert::{From, TryFrom};
 use std::fs::File;
-use std::io::Read;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


johanneskoester (changes requested):
Nice work! Can you do the generalization of the nomenclature in this PR as well? E.g. instead of somatic_tumor_events rather mutational_events. Instead of tumor mutational burden just mutational burden, etc.

PR #164, opened 2021-05-14
implement BDD based simplification, and automatic referral of events

johanneskoester:


src/errors.rs
```diff
@@ -60,4 +60,6 @@ pub(crate) enum Error {
     InvalidStrandInfo { value: char },
     #[error("invalid read orientation information '{value}', must be 'F1R2', 'F2R1', etc.")]
     InvalidReadOrientationInfo { value: String },
+    #[error("event {name} is unsatisfiabe: check whether it is defined as intended, if you are sure, it should be removed because it can never become true. This can e.g. happen if the event is a negation of other events that span the entire set of possibilites.")]
```
tedil, line 63:
```suggestion
    #[error("event {name} is unsatisfiable: check whether it is defined as intended, if you are sure, it should be removed because it can never become true. This can e.g. happen if the event is a negation of other events that span the entire set of possibilites.")]
```

PR #171, opened 2021-05-25
decode_phred: ensure quoting of description in changed header lines

dlaehnemann:
Otherwise, descriptions containing commas will lead to invalid header lines, as commas will be interpreted as field separators in the header line.


tedil (approved):
Nice catch!

PR #175, opened 2021-06-01
Haplotype call abundances

johanneskoester:


src/cli.rs
```diff
@@ -407,6 +407,29 @@ pub enum CallKind {
         )]
         output: Option<PathBuf>,
     },
+    // my part starts here.
+    #[structopt(
+        name = "haplotype-abundances",
+        about = "Call variants in haplotypes.",
```
johanneskoester, line 413:
```suggestion
        about = "Call haplotype abundances (e.g. for HLA typing or viral strain quantification).",
```

src/cli.rs
```diff
@@ -407,6 +407,29 @@ pub enum CallKind {
...
+                required = true,
+                help = "HDF5 haplotype counts calculated by Kallisto.",
+            )]
+            haplotype_counts: PathBuf,
+
+            #[structopt(
+                parse(from_os_str),
+                long = "haplotype-variants",
+                required = true,
+                help = "Haplotype variants should be supplied.",
```
johanneskoester, line 429:
```suggestion
                help = "Haplotype variants compared to a common reference.", // TODO later, we will add a subcommand to generate this file with Varlociraptor as well
```

test.txt
```diff
@@ -0,0 +1 @@
+this is a test.
```
johanneskoester, line 1: I guess this can be removed?

johanneskoester:
Also make sure to format your code with `cargo fmt` in the repo folder.

src/calling/haplotype_abundances.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[derive(Builder)]
+#[builder(pattern = "owned")]
+pub(crate) struct Caller<'a> {
+    hdf5_reader: hdf5::Reader<'a>,
+    vcf_reader: bcf::Reader,
+}
+
+impl<'a> Caller<'a> {
+    pub(crate) fn new<P: AsRef<Path>>(hdf5_path: P, vcf_path: P) -> Result<Self> {
```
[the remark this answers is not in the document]
huzuner: I think I fixed it now.

src/calling/haplotype_abundances.rs
```diff
@@ -1,43 +1,38 @@
...
-pub(crate) struct Caller<'a> {
-    hdf5_reader: hdf5::Reader<'a>,
+pub(crate) struct Caller {
+    hdf5_reader: hdf5::File,
     vcf_reader: bcf::Reader,
 }
 
-impl<'a> Caller<'a> {
+impl Caller {
+    pub(crate) fn cdf(&self) -> Result<()> {
```
johanneskoester, line 16: Good! I would expect this to have a parameter for the sequence name though. With regular kallisto usage, it would be the transcript id, in our case it would be the name of the respective contig (e.g. the HLA type, or the virus strain).

We will call this function for each of them.

src/calling/haplotype_abundances.rs
```diff
@@ -1,43 +1,38 @@
...
+        let mut bootstraps = Vec::new();
+        for i in 0..num_bootstraps {
+            let dataset = group.dataset(&format!("bs{i}", i = i))?;
+            let tpm = dataset.read_1d::<u64>()?;
+            bootstraps.extend(tpm.to_vec());
+        }
+        let n_values = hdf5.dataset("/aux/lengths")?.read_scalar()?;
+        let array = ndarray::Array2::from_shape_vec((num_bootstraps, n_values), bootstraps)?;
+        let array = array.t();
+        let cdf: Vec<_> = array
```
johanneskoester, line 29: It would be good to not only return the CDF, but also via a struct a second field containing the raw bootstrap values. This way we can plot them for QC. THis would also be the next task: add a command line flag to your subcommand that instructs varlociraptor to store plots for each quantified haplotype in a given directory. Such plotting is in rust best done via [vega-lite](https://vega.github.io/vega-lite). @jafors is an expert in that. Basically, vega-lite is a way to define a plot along with its data in JSON format. You basically just store that format from rust, and then subsequent rules in a workflow can render that to an SVG or PDF file using the vega-lite command line utility.

Regarding the plot:
x-axis: abundance
y-axis: density

The plot should superimpose the CDF with a histogram of the bootstrap values.

src/calling/haplotype_abundances.rs
huzuner: I did some changes, could you have a look at it again?

src/calling/haplotype_abundances.rs
```diff
@@ -3,22 +3,47 @@ use derive_builder::Builder;
...
-    pub(crate) fn cdf(&self, seqname: String) -> Result<ECDF> {
+impl Caller {  
+    //Below function outputs a vector of filtered seqnames according to min_tpm.
+    pub(crate) fn filter_seqnames(&self) -> Result<Vec<String>> {
+        let hdf5 = &self.hdf5_reader;
+        let tpm = hdf5.dataset("bootstrap/bs0")?.read_1d::<u64>()?;
+        let mut indices = Vec::new();
+        for (i,num) in tpm.iter().enumerate() {
+            if num > &self.min_tpm {
+                indices.push(i);
```
johanneskoester, line 31: I think the filtering should not happen based on the bootstrap TPMs, but rather based on the final estimate of kallisto (i.e. the maximum likelihood TPM).

src/calling/haplotype_abundances.rs
```diff
@@ -29,18 +54,37 @@ impl Caller {
...
     }
 
     pub(crate) fn call(&self) {
         todo!()
     }
 }
+
+impl ECDF {
+    pub(crate) fn qc(&self) -> Result<()> {
+        let blueprint = "../../templates/plots/qc_cdf.json";
```
johanneskoester, line 77: Maybe call the method `plot_qc`?

src/calling/haplotype_abundances.rs
```diff
@@ -29,18 +54,37 @@ impl Caller {
...
+        let mut blueprint: serde_json::Value = serde_json::from_str(blueprint)?;
+
+        let bootstraps = &self.bootstraps;
+        let ecdf = &self.ecdf;
+
+        let data = json!([bootstraps, ecdf]);
+        blueprint["data"]["values"] = data;
+
+        println!("{}", serde_json::to_string_pretty(&blueprint)?);
```
johanneskoester, line 86: There should be a separate JSON for each sequence. Hence, we cannot print them here, but they should be stored inside a given directory.

johanneskoester (commented):
Nice progress. See my comments below.

src/calling/haplotype_abundances.rs
johanneskoester: Ok, now we use the est_counts. This is already nice. However, they are not normalized for length. Is there a single TPM estimate we could use?

src/calling/haplotype_abundances.rs
```diff
@@ -0,0 +1,92 @@
...
+    
+    }
+
+    pub(crate) fn call(&self) {
+        todo!()
+    }
+}
+
+impl ECDF {
+    pub(crate) fn plot_qc(&self, qc_plot: Option<PathBuf> ) -> Result<()> {
```
johanneskoester, line 79: Looks good. Can you now connect this with the CLI, such that it is actually called? I'd like to see the CDFs on our simulated data now.

src/calling/haplotype_abundances.rs
```diff
@@ -85,6 +88,49 @@ impl Caller {
...
+            / count as f64;
+        let std = variance.sqrt();
+
+        let theta = std / mean;
+
+        let mle_dataset = hdf5.dataset("est_counts")?.read_1d::<f64>()?;
+        let mle_norm = mle_dataset / &seq_length; //normalized mle counts by length
+        let mle = mle_norm[index];
+
+        let neg_bin = neg_binom(mle, mean, theta);
```
johanneskoester, line 130: The code looks correct. Now it is time to use/move this in the right context: we can start with implementing the actual prediction now.

1. the Caller struct should store mle, mean and theta for each haplotype. You can do this in a HashMap, which is initially empty and filled by calling a new method `estimate_parameters`. THis method should contain the code above in which you calculate mle, mean, and theta from the hdf5.
2. Then, we need to build a Bayesian model on top of this, which calculates the posterior probability of each combination of haplotype fractions. There is a framework for this in rust-bio: https://docs.rs/bio/0.37.1/bio/stats/bayesian/model/index.html. I suggest though that we get this particular part of the implementation started together. This should help you a bit to feel more comfortable with this kind of stuff. How about Sep 15, 9:30?
3. Once we have the posterior probability for each haplotype fraction, we can, for now, output each haplotype with its maximum a posteriori fraction.

src/calling/haplotypes.rs
```diff
@@ -1,48 +1,85 @@
...
+        ])?; //depends upon the number of haplotypes
+
+        //write best record on top
+        let mut first_rec = Vec::new();
+        let best = posterior.next().unwrap();
+        let best_density = best.1 .0.exp();
+        let best_odds = 1;
+        first_rec.push(best_density.to_string());
+        first_rec.push(best_odds.to_string());
+        for j in 0..3 {
```
johanneskoester, line 62: I guess this hardcoded 3 has to be removed?

src/calling/haplotypes.rs
```diff
@@ -1,48 +1,85 @@
...
+            //the number depends upon the number of haplotypes
+            let fractions = best.0;
+            first_rec.push(fractions[j].to_string());
+        }
+        wtr.write_record(first_rec)?;
+
+        //write the rest of the records
+        for i in posterior {
```
johanneskoester, line 70: i is not a good variable name here. Use something speaking.

src/calling/haplotypes.rs
```diff
@@ -1,48 +1,85 @@
...
+            let mut other_rec = Vec::new();
+            let density = i.1 .0.exp();
```
johanneskoester, line 72: I'd rather use "unpacking" instead of index access with tuples. E.g. in the loop above you can have:
```
for (haplotype_frequencies, prob) in posterior {
```

src/calling/haplotypes.rs
```diff
@@ -1,48 +1,85 @@
...
+        wtr.write_record(&[
+            "posterior_prob(log)",
+            "odds ratio",
+            "haplotype_a",
+            "haplotype_b",
+            "haplotype_c",
+        ])?; //depends upon the number of haplotypes
+
+        //write best record on top
+        let mut first_rec = Vec::new();
```
johanneskoester, line 56: Instead of composing a vector with the entries here, it is more idiomatic to define a struct with the respective fields and use the serialization capabilities of the csv crate: https://docs.rs/csv/1.1.6/csv/#example-with-serde

src/calling/haplotypes.rs
```diff
@@ -1,48 +1,85 @@
...
+            let odds = density / best_density;
+            other_rec.push(density.to_string());
+            other_rec.push(odds.to_string());
+            for j in 0..3 {
+                let fractions = i.0;
+                other_rec.push(fractions[j].to_string());
+            }
+            wtr.write_record(other_rec)?;
+        }
+        wtr.flush()?;
```
johanneskoester, line 82: No need to flush, this happens automatically when the object goes out of scope.

src/haplotypes/model.rs
```diff
@@ -98,15 +99,19 @@ impl Likelihood {
...
-        //     .0
-        //     .iter()
-        //     .zip(data.kallisto_estimates.iter())
-        //     .map(|(fraction, estimate)| {
-        //         Self::neg_binom(estimate.count, fraction, estimate.dispersion)
-        //     });
-        todo!()
+        //Later: use the cache to avoid redundant computations.
+        event
+            .0
```
johanneskoester, line 104: This `.0`looks fishy to me. At least I don't understand why that is done here. Either comment unpack the event or iterate over all entries. Whatever is appropriate here.

src/calling/haplotypes.rs
```diff
@@ -22,6 +22,15 @@ pub struct Caller {
     outcsv: Option<PathBuf>,
 }
 
+#[derive(Serialize)]
+struct Output {
+    haplotype_a: NotNaN<f64>,
+    haplotype_b: NotNaN<f64>,
+    haplotype_c: NotNaN<f64>,
```
johanneskoester, line 29: But we don't always have 3 haplotypes only. I think it would make most sense to use a hash map here. The problem is that I think the csv crate does not support the serialization of hashmaps or other kinds of maps yet. Hence, I'm sorry but I guess my suggestion was not the right idea. Instead, you should probably just write the records as Vec of the values in the right order.

huzuner: No problem, I will do that.

src/calling/haplotypes.rs
```diff
@@ -53,36 +43,39 @@ impl Caller {
         // with each column after the first showing the fraction of the respective haplotype
         let mut posterior = m.event_posteriors();
         let mut wtr = csv::Writer::from_path(self.outcsv.as_ref().unwrap())?;
-        let mut haplotype_fractions = Vec::new();
+        wtr.write_record(&[
+            "density",
+            "odds",
+            "haplotype_a",
+            "haplotype_b",
+            "haplotype_c",
```
johanneskoester, line 51: Ok, and here one should not put hardcoded names, but instead build the haplotype columns from the input (i.e. the names coming from the kallisto data).

src/calling/haplotypes.rs
```diff
@@ -34,22 +34,24 @@ impl Caller {
 
         // Step 3: calculate posteriors.
         //let m = model.compute(universe, &data);
-        let n_of_haplotypes = 3;
-        let m = model.compute_from_marginal(&Marginal::new(n_of_haplotypes), &data);
+        //let n_of_haplotypes = 3;
+        let mut haplotypes = Vec::new();
```
johanneskoester, line 38: Just use `let haplotypes: Vec<_> = kallisto_estimates.keys().collect();`

src/calling/haplotypes.rs
```diff
@@ -34,22 +34,24 @@ impl Caller {
...
-            "density",
-            "odds",
-            "haplotype_a",
-            "haplotype_b",
-            "haplotype_c",
-        ])?; //depends upon the number of haplotypes
+        let mut headers = vec!["density", "odds"];
+        for h in &haplotypes {
+            headers.push(h);   
+        }
```
johanneskoester, line 53: Here use `headers.extend(&haplotypes)`;

src/calling/haplotypes.rs
```diff
@@ -0,0 +1,176 @@
...
+        wtr.write_record(&headers)?; //depends upon the number of haplotypes
+
+        //write best record on top
+        let mut records = Vec::new();
+        let best = posterior.next().unwrap();
+        let best_density = best.1 .0.exp();
```
johanneskoester, line 59: Rather write something like
```
let (haplotype_frequencies, best_density) = posterior.next().unwrap();
```

src/calling/haplotypes.rs
```diff
@@ -0,0 +1,176 @@
...
+        for haplotype in 0..haplotypes.len() {
+            let haplotype_frequencies = best.0;
+            records.push(haplotype_frequencies[haplotype].to_string());
+        }
+        wtr.write_record(records)?;
+
+        //write the rest of the records
+        for (haplotype_frequencies, logprob) in posterior {
+            let mut records = Vec::new();
+            let density = logprob.0.exp();
```
johanneskoester, line 73:
```suggestion
            let density = logprob.exp();
```

src/calling/haplotypes.rs
```diff
@@ -34,11 +34,7 @@ impl Caller {
 
         // Step 3: calculate posteriors.
         //let m = model.compute(universe, &data);
-        //let n_of_haplotypes = 3;
-        let mut haplotypes = Vec::new();
-        for key in kallisto_estimates.keys() {
-            haplotypes.push(key);
-        }
+        let haplotypes: Vec<_> = kallisto_estimates.keys().map(|x| x.to_string()).collect();
```
johanneskoester, line 37: Instead of the map(), you can use cloned(), which has the same effect. No need to write the map down manually.

src/calling/haplotypes.rs
```diff
@@ -47,30 +43,28 @@ impl Caller {
...
+        let mut headers: Vec<_> = vec!["density".to_string(), "odds".to_string()];
+        headers.extend(haplotypes);
+        wtr.write_record(&headers)?;
 
         //write best record on top
         let mut records = Vec::new();
         let best = posterior.next().unwrap();
-        let best_density = best.1 .0.exp();
+        let (haplotype_frequencies, best_density) = posterior.next().unwrap();
+        let best_density = best_density.exp();
```
johanneskoester, line 54: No need to do the exp here.

src/calling/haplotypes.rs
```diff
@@ -47,30 +43,28 @@ impl Caller {
...
+            records.push(frequency.to_string());
         }
         wtr.write_record(records)?;
 
         //write the rest of the records
-        for (haplotype_frequencies, logprob) in posterior {
+        for (haplotype_frequencies, density) in posterior {
             let mut records = Vec::new();
-            let density = logprob.0.exp();
+            let density = density.exp();
```
johanneskoester, line 67: No need for exp here. Keep density and best_density in log space, and calculate odds as

```
let odds = (density - best_density).exp();
records.push(density.exp().to_string());
records.push(odds.to_string());
```

The reason is that this will be more precise for very small densities (which can otherwise be just interpreted as 0 when doing exp, and then we could have e.g. a division by zero or an odds of zero).

src/calling/haplotypes.rs
huzuner: But the extend method is only accepting the same types to be extended. That's why I needed to have the `to_string`.

src/calling/haplotypes.rs
```diff
@@ -166,3 +172,54 @@ impl KallistoEstimates {
...
+pub(crate) struct Variant(#[deref] String);
+
+#[derive(Derefable, Debug, Clone, PartialEq, Eq)]
+pub(crate) struct HaplotypeVariants(#[deref] HashMap<Variant, Vec<Haplotype>>);
+
+impl HaplotypeVariants {
+    pub(crate) fn new(haplotype_variants: &mut bcf::Reader) -> Result<Vec<HaplotypeVariants>> {
+        let mut vec_haplotype_variants = Vec::new();
+        for record_result in haplotype_variants.records() {
+            let record = record_result.expect("Failed to read record");
```
johanneskoester, line 186: Use `?`instead of expect, this returns the error as a result.

src/calling/haplotypes.rs
```diff
@@ -166,3 +172,54 @@ impl KallistoEstimates {
...
+            }
+
+            //store the haplotypes that carry the variant
+            let header = record.header();
+            let gts = record.genotypes().expect("Error reading genotypes"); //genotypes of all samples
+            let sample_count = usize::try_from(record.sample_count()).unwrap();
+            let sample = (0..sample_count).collect::<Vec<_>>();
+            let mut haplotypes: Vec<Haplotype> = Vec::new();
+
+            for (sample_index, mut x) in sample.iter().zip(header.samples().into_iter()) {
```
johanneskoester, line 207: To get the index of each sample/haplotype, just use `enumerate()` on the iterator over the `samples()`.

src/calling/haplotypes.rs
```diff
@@ -166,3 +172,54 @@ impl KallistoEstimates {
...
+            for allele in record.alleles() {
+                for c in allele {
+                    let base = char::from(*c).to_string();
+                    variant.push_str(&format!("{}{}", base, "_"));
+                }
+            }
+
+            //store the haplotypes that carry the variant
+            let header = record.header();
+            let gts = record.genotypes().expect("Error reading genotypes"); //genotypes of all samples
```
johanneskoester, line 202: Return error via `?` instead of using expect.

src/calling/haplotypes.rs
```diff
@@ -166,3 +172,54 @@ impl KallistoEstimates {
...
+            let sample_count = usize::try_from(record.sample_count()).unwrap();
+            let sample = (0..sample_count).collect::<Vec<_>>();
+            let mut haplotypes: Vec<Haplotype> = Vec::new();
```
johanneskoester, line 205: This wastes too much space. I'd rather use a [bitvector](https://docs.rs/bv), with a bit for each haplotype, and insert a 1 into the bitvector if the corresponding haplotype has the variant.
The haplotype index should thereby come from whatever you use in the rest of the code (order given at the command line perhaps?

johanneskoester (commented):
Good start! See below.

src/calling/haplotypes.rs
```diff
@@ -177,49 +184,93 @@ impl KallistoEstimates {
...
+                    }
+                } else {
+                    //the case for structural variants
+                    for c in allele {
+                        let base = char::from(*c).to_string();
+                        variant.push_str(&base);
+                    }
+                    variant.push_str("_")
                 }
             }
```
johanneskoester, line 218: I don't think we need to store variants like this. Granted you need a correspondence between the BCF with the varlociraptor calls and the BCF with the variant to haplotype assignment. But instead of building a key for this from scratch like you do here, it is better to use `record.id()`. For this to work, you just need to ensure that your variant to haplotype BCF contains a unique ID for each row (in the ID column of the BCF). This can be done during generation of that file (e.g. just increment a number).

src/calling/haplotypes.rs
```diff
@@ -177,49 +184,93 @@ impl KallistoEstimates {
...
+                for gta in gts.get(sample_index).iter() {
                     if gta == &Unphased(1) {
-                        let mut s = String::new();
-                        x.read_to_string(&mut s);
-                        haplotypes.push(Haplotype(s));
+                        bv.push(true);
                     } else {
+                        bv.push(false);
                         break;
                     }
```
johanneskoester, line 240: Simpler, more idiomatic version:
```
bv.push(*gta == Unphased(1))
```

src/calling/haplotypes.rs
```diff
@@ -177,49 +184,93 @@ impl KallistoEstimates {
...
-            let sample_count = usize::try_from(record.sample_count()).unwrap();
-            let sample = (0..sample_count).collect::<Vec<_>>();
-            let mut haplotypes: Vec<Haplotype> = Vec::new();
+            let gts = record.genotypes()?; //genotypes of all samples
 
-            for (sample_index, mut x) in sample.iter().zip(header.samples().into_iter()) {
-                for gta in gts.get(*sample_index).iter() {
+            let mut bv: BitVec<usize> = BitVec::new();
+            for (sample_index, x) in header.samples().into_iter().enumerate() {
+                for gta in gts.get(sample_index).iter() {
```
johanneskoester: I am unsure whether this will give you the correct order in the resulting bitvector. If I get your decision above right, you chose to take the kallisto file as the reference order for the haplotypes. Then, I think here you will have to ensure that the bitvector is filled in the same order. Otherwise you won't be able to use the same index later on to access both data.

src/calling/haplotypes.rs
```diff
@@ -177,49 +184,93 @@ impl KallistoEstimates {
...
+        let mut calls = HashMap::new();
+        for record_result in haplotype_calls.records() {
+            let record = record_result?;
+
+            //store the variant and the afd if the prob_present field is not nan (to save space)
+            if !record.info(b"PROB_PRESENT").float()?.unwrap()[0].is_nan() {
+                let mut variant = String::new();
+                variant.push_str(&format!("{}{}", record.contig(), "_"));
+                let pos = record.pos().to_string();
+                variant.push_str(&format!("{}{}", pos, "_"));
```
johanneskoester, line 263: See my comment above, let us use `record.id()` to identify the variant and match it between the assignment BCF and the calls here.

src/calling/haplotypes.rs
```diff
@@ -177,49 +184,93 @@ impl KallistoEstimates {
...
+                            variant.push_str(&base);
+                        }
+                        variant.push_str("_")
+                    }
+                }
+
+                //store the afd string
+                let afd_utf = record.format(b"AFD").string()?[0];
+                let afd = std::str::from_utf8(afd_utf).unwrap();
+                calls.insert(Variant(variant.clone()), afd.to_string());
```
johanneskoester, line 283: I don't think we should keep the string here. Rather store this as a `HashMap<VariantId, HashMap<AlleleFreq, LogProb>>` here (with VariantId being the ID you get from record.id()). This way, you avoid recomputing this in the model methods, where it would be unnecessarily expensive because they are evaluated multiple times.

src/haplotypes/model.rs
```diff
@@ -124,7 +126,41 @@ impl Likelihood {
     ) -> LogProb {
         // TODO compute likelihood based on Varlociraptor VAFs.
         // Let us postpone this until we have a working version with kallisto only.
-        LogProb::ln_one()
+        let variant_matrix: Vec<BitVec> = data.haplotype_variants.values().cloned().collect();
+        let likelihood = Prob(1.0);
+        for (variant, bv) in data.haplotype_variants.iter() {
+            let variant_sum = 0.0;
+            for (i, fraction) in event.iter().enumerate() {
+                if bv[i.try_into().unwrap()] == true {
```
johanneskoester, line 134: Use `zip` here to iterate over event and bitvector at the same time.

src/haplotypes/model.rs
```diff
@@ -124,7 +126,41 @@ impl Likelihood {
...
+                    let variant_sum = variant_sum + NotNaN::into_inner(*fraction);
```
johanneskoester, line 135: Instead of doing this summing manually in a loop, make use of Rusts iterator methods:
```
let variant_sum: AlleleFreq = event.iter().zip(bv.iter()).filter_map(|(fraction, is_present)| {
    if is_present {
        Some(*fraction)
    } else {
        None
    }
}).sum();
```

src/haplotypes/model.rs
```diff
@@ -124,7 +126,41 @@ impl Likelihood {
...
+            let mut densities = Vec::new();
+            if data.haplotype_calls.contains_key(variant) {
+                let afd = data.haplotype_calls[variant].split(",");
+                for pair in afd {
+                    let (vaf, density) = pair.split_once("=").unwrap();
+                    let (vaf, density): (f64, f64) =
+                        (vaf.parse().unwrap(), density.parse().unwrap());
+                    vafs.push(vaf);
+                    densities.push(density);
+                }
```
johanneskoester, line 148: As mentioned above, move the parsing into the place where you initialize the HaplotypeCalls struct.

src/haplotypes/model.rs
```diff
@@ -124,7 +126,41 @@ impl Likelihood {
...
+                    let vaf_sum = f64::from(Prob::from(PHREDProb(variant_sum)));
+                    if vaf == &vaf_sum {
+                        let likelihood = likelihood * Prob::from(PHREDProb(*density));
+                    } else {
+                        let density = densities[i - 1]
+                            + (vaf_sum - vafs[i - 1]) * (densities[i + 1] - densities[i - 1])
+                                / (vafs[i + 1] - vafs[i - 1]);
+                        let likelihood = likelihood * Prob::from(PHREDProb(density));
+                    }
+                }
```
johanneskoester, line 159: This goes in the right direction, but is not ideal yet because it is slow. I have above suggested to use a HashMap for storing the densities. Let me revise that, better use a BTreeMap. Then, you can actually get either the exact matching entry or those two surrounding your desired VAF at logarithmic time, with API calls instead of the need to write a loop here. And the code for that kind of lookup should be put as a method into the HaplotypeCalls struct (the paradigm that applies here is "separation of concerns").

src/haplotypes/model.rs
huzuner: As far as I understand, `bv::BitVec` does not implement the Iterator trait:https://docs.rs/bv/0.10.0/bv/struct.BitVec.html

src/haplotypes/model.rs
huzuner: As far as I understand, bv::BitVec does not implement the Iterator trait:https://docs.rs/bv/0.10.0/bv/struct.BitVec.html. Do you have some other suggestion?

src/haplotypes/model.rs
```diff
@@ -127,37 +132,16 @@ impl Likelihood {
         // TODO compute likelihood based on Varlociraptor VAFs.
         // Let us postpone this until we have a working version with kallisto only.
         let variant_matrix: Vec<BitVec> = data.haplotype_variants.values().cloned().collect();
-        let likelihood = Prob(1.0);
-        for (variant, bv) in data.haplotype_variants.iter() {
+        let mut likelihood = Prob(1.0);
+        for (bv, (_, afd)) in variant_matrix.iter().zip(data.haplotype_calls.iter()) {
             let variant_sum = 0.0;
             for (i, fraction) in event.iter().enumerate() {
                 if bv[i.try_into().unwrap()] == true {
```
johanneskoester, line 139: `== true` is never needed in an if condition, since it already checks for the condition to be true anyway.

src/haplotypes/model.rs
```diff
@@ -127,37 +132,16 @@ impl Likelihood {
         // TODO compute likelihood based on Varlociraptor VAFs.
         // Let us postpone this until we have a working version with kallisto only.
         let variant_matrix: Vec<BitVec> = data.haplotype_variants.values().cloned().collect();
-        let likelihood = Prob(1.0);
-        for (variant, bv) in data.haplotype_variants.iter() {
+        let mut likelihood = Prob(1.0);
+        for (bv, (_, afd)) in variant_matrix.iter().zip(data.haplotype_calls.iter()) {
```
johanneskoester, line 136: `bv` is not a good name. The name should rather tell something about the meaning, not the type of the variable.

johanneskoester: Also, instead of having a for loop here, it is more idiomatic to use something like
```
let likelihood = variant_matrix.iter().zip(data.haplotype_calls.iter()).map(| ... | {
    ...
}).sum()
```

src/haplotypes/model.rs
```diff
@@ -127,37 +132,16 @@ impl Likelihood {
...
-                    } else {
-                        let density = densities[i - 1]
-                            + (vaf_sum - vafs[i - 1]) * (densities[i + 1] - densities[i - 1])
-                                / (vafs[i + 1] - vafs[i - 1]);
-                        let likelihood = likelihood * Prob::from(PHREDProb(density));
-                    }
+                    let variant_sum = variant_sum + NotNan::into_inner(*fraction);
                 }
             }
+            let vaf_sum = NotNan::new(f64::from(Prob::from(PHREDProb(variant_sum)))).unwrap();
```
johanneskoester, line 143: Why all these conversions? The fraction is an AlleleFreq, which can simply be summed up.

src/calling/haplotypes.rs
```diff
@@ -225,62 +210,67 @@ impl HaplotypeVariants {
...
+        if self.contains_key(&vaf) {
+            let likelihood = likelihood * Prob::from(PHREDProb(*self.get(&vaf).unwrap()));
+        } else {
+            let (x_0, y_0) = self.range(..vaf).next_back().unwrap();
+            let (x_1, y_1) = self.range(vaf..).next().unwrap();
+
+            let density = NotNan::new(*y_0).unwrap() + (vaf - *x_0) * (*y_1 - *y_0) / (*x_1 - *x_0); //calculation of density for given vaf by linear interpolation
+            let likelihood = likelihood * Prob::from(PHREDProb(NotNan::into_inner(density)));
+        }
+        Ok(*likelihood)
```
johanneskoester, line 243: Here you just return the likelihood that was given as an argument. But instead, there should be no such argument and you should return what you have computed in the if-else.

src/calling/haplotypes.rs
```diff
@@ -225,62 +210,67 @@ impl HaplotypeVariants {
                 let mut s = String::new();
                 x.read_to_string(&mut s);
                 if haplotypes.contains(&s) {
-                    haplotype_indices.push(sample_index);
+                    haplotype_indices.push(sample_index); //pushing the exact indices of haplotypes that are found in header.samples()
```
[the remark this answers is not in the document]
huzuner: I have added a new commit to perform your suggestion here: https://github.com/varlociraptor/varlociraptor/blob/9b7204bc03338f230ef16c2d5b9763e3d7d159a9/src/calling/haplotypes.rs#L218. Which operation would you think we should keep to collect the haplotype indices?

PR #176, opened 2021-06-02
feat: checking for overlapping events in scenario.yaml

tedil:
This draft PR adds a little routine for checking whether events defined in scenario.yaml overlap. However, I am unsure how exactly contig-specific negations work, which is why I have opted to use contig="all" for now.

And since `Formula::normalize` returns `NormalizedFormula`, it wasn't easily possible to build the disjunction and re-normalize, so for the time being there's `normalize2` which returns `Formula` instead.

This works for discrete events, not sure how intervals are modeled internally.

I'm also not quite happy about the return type (I first thought that `bool` / `Result<bool>` should be enough, but then error messages couldn't be particularly helpful).


tedil:
While investigating interval overlaps, I noticed that events such as
`baz: "normal:[0.0,0.7]&normal:[0.3,1.0]"` normalize to `normal:[0.0,01]`, instead of `normal:[0.3,0.7]` instead.
I suspect this is due to cases `VAFRangeOverlap::End` and `VAFRangeOverlap::Start` being swapped.
Also, disjunctions of intervals do not get merged at all at the moment, even though they could be merged *if* they overlap.

johanneskoester:
If you want, you can fix that in the PR.

src/grammar/mod.rs
```diff
@@ -216,6 +217,49 @@ impl Scenario {
...
             .collect()
     }
+
+    pub(crate) fn validate(&self) -> Result<()> {
+        let expressions = self
+            .expressions
+            .iter()
+            .filter(|(id, _)| id.0 != "absent")
+            .map(|(_, expr)| expr.normalize2(self))
+            .collect::<Result<HashSet<_>>>()?;
```
johanneskoester, line 227: Usually it is easier on the eye to put the type behind the let.

src/grammar/mod.rs
```diff
@@ -216,6 +217,49 @@ impl Scenario {
             })
             .collect()
     }
+
+    pub(crate) fn validate(&self) -> Result<()> {
```
johanneskoester, line 221: I think it is safer to do this when the contig is known. This would be in the vaftrees function then. This happens only once per contig, so the overhead is minimal.

johanneskoester: Then, we also do not need another version of the normalize function.

johanneskoester (changes requested):

src/grammar/mod.rs
tedil: Thanks, I didn't really know where the proper entrypoint for that is

src/grammar/mod.rs
tedil: Yes but that won't work in this case, since I'm collapsing/flattening the `Result<Formula>` items of the Iterator (such that there's only one result at the end) while using the `?` operator at the same time.

src/grammar/mod.rs
tedil: > Then, we also do not need another version of the normalize function.

Hm, but `normalize` returns type `NormalizedFormula`, which is problematic since we have two `NormalizedFormula`s, for which we want to check if their disjunction exists in the event definitions, but we need to normalize that disjunction, but `normalize` is only defined for `Formula`, not `NormalizedFormula`.

Perhaps instead of returning `NormalizedFormula`, return `T: Into<NormalizedFormula>` instead (and have `NormalizedFormula` implement `From<Formula>`)?
Or the other way around, i.e. make `Formula` implement `From<NormalizedFormula>` so that normalized formulas can be used as ordinary formulas again. Probably this.

tedil: Also `vaftrees` doesn't get called if there's no rid set (for example when there are no records to be processed I guess?) not quite sure of the inner workings here. That'd also make scenario validation data dependent in some sense, I'd much rather have it validated before anything happens, i.e. upon parsing the scenario from the yaml. Though I'm not sure where to get the contigs from, then.

johanneskoester: > Also `vaftrees` doesn't get called if there's no rid set (for example when there are no records to be processed I guess?) not quite sure of the inner workings here. That'd also make scenario validation data dependent in some sense, I'd much rather have it validated before anything happens, i.e. upon parsing the scenario from the yaml. Though I'm not sure where to get the contigs from, then.

True, but if there is no data, it also does not really matter. I know what you mean, but I think we can live with that little caveat.

src/grammar/mod.rs
tedil: Have tried different constellations, this is the one with the fewest characters and maintaining okay-ish readability

johanneskoester:
Do you have an estimate when this will be ready for merging?

johanneskoester:
Closing in favor of #177.

PR #177, opened 2021-06-04
fix: merging of VAFRanges

tedil:
The merge of two `VAFRange`s in `FormulaTerminal` which was supposed to result in the intersection (i.e. conjunction) of the two ranges but actually resulted in the union (i.e. disjunction). 
This PR fixes that behaviour for merging conjunctions.
However, there may also be certain disjunction cases (such as overlapping ranges) which ought to be merged when normalizing, e.g. `normal:[0.0, 0.5] | normal:[0.5, 1.0]` can be merged to `normal[0.0,1.0]`. I'll hence leave this as a draft for now.


tedil:
> However, there may also be certain disjunction cases (such as overlapping ranges) which ought to be merged when normalizing, e.g. `normal:[0.0, 0.5] | normal:[0.5, 1.0]` can be merged to `normal[0.0,1.0]`. I'll hence leave this as a draft for now.

Ah, but while intersections can be done one after the other, unions have to be done at the same time *after* all other normalizations have been done, so that expressions such as: `(normal:[0.0, 0.25] | normal:[0.5,0.75]) | (normal:[0.25,0.5] | normal:[0.75,1.0])` can actually be merged to `normal:[0.0,1.0]`.

johanneskoester:
Do you have an estimate when this could be ready for merging?

tedil:
> Do you have an estimate when this could be ready for merging?

What I think needs to be done before merging: find out why test 71 fails (because of a terminal `false` I guess?)
What I think *should* be done before merging: implement merging range unions (see `(normal:[0.0, 0.25] | normal:[0.5,0.75]) | (normal:[0.25,0.5] | normal:[0.75,1.0])` example in comment further up).

I think I can address the first point tomorrow, but not sure how long the second one takes, kind of depends on how exactly BDDs are used for the formulas in varlociraptor (I assume it treats VAFSpectrum Terminals simply as symbols, i.e. distinct variables and no further interpretation happens).

tedil:
While this cannot cover all cases, I implemented a greedy approach to merging range disjunctions: `VAFRange`s/`VAFSet`s get sorted by their start position and subsequently merged as long as possible. I now realize I should've just built a tree.

tedil:
I think this is ready for review. However, this PR has:
1. the "intersection instead of union for conjunctions" fix (see initial comment)
2. "validate scenarios by checking for duplicate events" which was originally intended for PR #176
3. "union for disjunctions of VAFSpectra"

johanneskoester (approved):
All I can say is awesome 👏!

PR #178, opened 2021-06-10
Scatter/Contour plot functionality

jafors:
A function to plot a scatter plot with an overlaying contour plot between two sample groups. Multiple samples can be plotted against one fixed sample.


src/estimation/sample_variants.rs
```diff
@@ -0,0 +1,168 @@
...
+    for s in tumor {
+        let tumor_id = bcf
+            .header()
+            .sample_id(s.as_bytes())
+            .unwrap_or_else(|| panic!("Sample {} not found", s));
+        tumor_ids.insert(s, tumor_id);
+    }
+
+    'records: loop {
+        let mut rec = bcf.empty_record();
```
johanneskoester, line 74: Just iterate over `bcf.records()`.

src/estimation/sample_variants.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    let mut tumor_ids = BTreeMap::new();
+
+    for s in tumor {
+        let tumor_id = bcf
+            .header()
+            .sample_id(s.as_bytes())
+            .unwrap_or_else(|| panic!("Sample {} not found", s));
+        tumor_ids.insert(s, tumor_id);
+    }
```
johanneskoester, line 71: Maybe do not talk about tumor/normal here but in general sample_a and sample_b?

src/cli.rs
```diff
@@ -293,6 +295,33 @@ pub enum PlotKind {
         #[structopt(long = "sample", required = true, help = "Sample to plot.")]
         sample: String,
     },
+    #[structopt(
+        name = "scatter",
+        about = "Plot variant allelic fraction scatter plot overlayed with a contour plot between two sample groups",
+        usage = "varlociraptor plot scatter --somatic-tumor-events SOMATIC_TUMOR \
+        --tumor-sample tumor --normal-sample normal < calls.bcf | vg2svg > scatter.svg",
```
johanneskoester, line 302: Can we have generic names instead of tumor/normal?

src/cli.rs
```diff
@@ -293,6 +295,33 @@ pub enum PlotKind {
...
+        setting = structopt::clap::AppSettings::ColoredHelp,
+    )]
+    Scatter {
+        #[structopt(
+            long = "somatic-tumor-events",
+            default_value = "SOMATIC_TUMOR",
```
johanneskoester, line 308: Here as well, keep it generic, not implying that this shall be only used for somatic variants. Although it of course makes most sense there, but nevertheless.

johanneskoester (changes requested):

johanneskoester:
Thanks a lot! One more thing: can you please also fix the clippy lints that occur within your new commits (not the old ones from elsewhere).

src/cli.rs
```diff
@@ -310,17 +310,17 @@ pub enum PlotKind {
         )]
         somatic_tumor_events: Vec<String>,
         #[structopt(
-            long = "normal-sample",
-            default_value = "normal",
-            help = "Name of the normal or reference sample in the given VCF/BCF."
+            long = "sample-A",
```
johanneskoester, line 313: Maybe just as positional arguments?

johanneskoester (changes requested):

johanneskoester (approved):

johanneskoester (changes requested):
Mhm, why are the alternative samples on the y-axis? Wouldn't it be more natural to have them on the y-axis?

jafors:
Yes, you are right. It looked "nicer" to me in the examples I tried. But that's probably just some weird bias and my own taste. Logically, it makes probably more sense the other way round.

johanneskoester (approved):

PR #187, opened 2021-07-07
fix: testcase 71 sometimes fails because formula normalization via BDD is not deterministic

tedil:
### Description

testcase 71 sometimes fails because formula normalization via BDD is not deterministic.

### QC


tedil:
We now skip checks whether the events already contain the disjunction of two events `A | B` if `A` or `B` are `False`. That at least fixes test 71. It does not, however, adequately answer the question if normalized formulas are always canonical (even though they should, since BDDs should ensure that).

johanneskoester (approved):

PR #198, opened 2021-08-11
Plot probabilites

FelixMoelder:
This adds functionality for plotting variant log probabilities vs chromosomal positions as a scatter plot.

### QC


src/variants/variant_probabilities.rs
```diff
@@ -0,0 +1,73 @@
...
+    pos: i64,
+}
+
+pub(crate) fn variant_probabilities() -> Result<()> {
+    let mut bcf = bcf::Reader::from_stdin()?;
+
+    let mut plot_data = Vec::new();
+    let header = bcf.header().to_owned();
+    for record in bcf.records() {
+        let rec = record.unwrap();
```
johanneskoester, line 22: Use error handling with ? here.

src/variants/variant_probabilities.rs
```diff
@@ -0,0 +1,73 @@
...
+        if let Some(rid) = rec.rid() {
+            let chr = String::from_utf8(header.rid2name(rid)?.to_vec())?;
+            let pos = rec.pos() - rec.rlen();
```
johanneskoester, line 25: Why?

src/variants/variant_probabilities.rs
```diff
@@ -0,0 +1,73 @@
...
+            let (event, prob) = max_prob(
+                (
+                    String::from("PROB_PRESENT"),
+                    rec.info(b"PROB_PRESENT").float().unwrap().unwrap()[0],
+                ),
+                (
+                    String::from("PROB_FFPE_ARTIFACT"),
+                    rec.info(b"PROB_FFPE_ARTIFACT").float().unwrap().unwrap()[0],
+                ),
+            );
```
johanneskoester, line 42: This shall not be specific for your scenario. The scatter plot below should always get the event with the highest probability. If that is ABSENT or ARTIFACT, the record should be skipped.

johanneskoester (changes requested):

PR #202, opened 2021-08-19
feat: added VAF log fold change and comparison operators to the scenario event grammar

johanneskoester:
### Description

This PR adds comparison expressions using `==`, `<`, `>`, `!=`, `<=`, `>=`for comparison between sample VAFs to the grammar, as well as an expression `lfc(sample_a, sample_b)` for defining constraints on the log fold change of VAFs between samples. 

### QC


src/utils/log2_fold_change.rs
```diff
@@ -0,0 +1,49 @@
...
+use ordered_float::NotNan;
+
+use crate::variants::model::AlleleFreq;
+use crate::utils::comparison::Comparison;
+
+#[derive(Debug, Clone)]
+pub(crate) struct Log2FoldChange {
+    vaf_a: AlleleFreq,
+    vaf_b: AlleleFreq,
+    l2fc: f64,
```
johanneskoester, line 12:
```suggestion
    value: f64,
```

src/utils/log2_fold_change.rs
```diff
@@ -0,0 +1,49 @@
...
+        Self {
+            vaf_a,
+            vaf_b,
+            l2fc: vaf_a.log2() - vaf_b.log2(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct Log2FoldChangePredicate(pub(crate) Comparison, pub(crate) NotNan<f64>);
```
johanneskoester, line 26: Rather use a struct and two fields `comp` and `value` here? Do we need Ord and PartialOrd (does not make sense to me)?

src/utils/log2_fold_change.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl Log2FoldChangePredicate {
+    pub(crate) fn is_true(&self, lfc: &Log2FoldChange) -> bool {
+        // FIXME: equality checks should probably allow for some error
+        let v = *self.1;
+        match self.0 {
+            Comparison::Equal => lfc.l2fc == v,
```
johanneskoester, line 33: For the strict comparison lints, there should indeed be taken seriously here. Just use `relative_eq!` from the approx crate (already a dependency).

src/utils/log2_fold_change.rs
tedil: Ord and PartialOrd are needed because this is a requirement for the some stuff further up the chain.

tests/resources/testcases/test_l2fc/scenario.yaml
```diff
@@ -0,0 +1,12 @@
+samples:
+  normal:
+    resolution: 10
+    universe: "0.0 | 0.5 | 1.0"
```
johanneskoester, line 4: For this testcase, I would have expected a continuous range in both samples, and also not consider them as tumor and normal but just a and b or so.

tedil: good point! I'll change that

src/utils/comparison.rs
```diff
@@ -10,17 +11,30 @@ pub enum Comparison {
     NotEqual,
 }
 
-impl Not for Comparison {
+impl ComparisonOperator {
+    pub(crate) fn is_true(&self, vaf_a: AlleleFreq, vaf_b: AlleleFreq) -> bool {
+        match self {
+            ComparisonOperator::Equal => relative_eq!(*vaf_a, *vaf_b),
```
tedil, line 17: Since we know the resolution of the universe, we can probably use a different (usually larger) epsilon for `relative_eq!` than the default.

tedil:
Since the L2FC operator always involves a comparison, we could integrate this into the Comparison path (e.g. by. adding a boolean flag field for that) or just leave it separate (as it is now, which incurs some code duplications)

tedil:
For `test_cmp` (where sample_a has VAF 1.0 at pos 380 and sample_b has VAF ~0.48 at pos 380), both MAP AFs are 0.74; this does not happen in `test_l2fc`, where the MAP AFs are 1.0 and 0.48 respectively. Perhaps `sample_a < sample_b` skips too many combinations?

johanneskoester:
> Since the L2FC operator always involves a comparison, we could integrate this into the Comparison path (e.g. by. adding a boolean flag field for that) or just leave it separate (as it is now, which incurs some code duplications)

Actually it was meant to be exactly like that. In my draft, I had implemented the L2FC path and expected you to translate a comparison in the grammar into an L2FC based formula.

johanneskoester:
> For `test_cmp` (where sample_a has VAF 1.0 at pos 380 and sample_b has VAF ~0.48 at pos 380), both MAP AFs are 0.74; this does not happen in `test_l2fc`, where the MAP AFs are 1.0 and 0.48 respectively. Perhaps `sample_a < sample_b` skips too many combinations?

Sounds like that. I guess this will be fixed by just using L2FC for everything, see my comment https://github.com/varlociraptor/varlociraptor/pull/202#issuecomment-903735847

tedil:
> > Since the L2FC operator always involves a comparison, we could integrate this into the Comparison path (e.g. by. adding a boolean flag field for that) or just leave it separate (as it is now, which incurs some code duplications)
> 
> Actually it was meant to be exactly like that. In my draft, I had implemented the L2FC path and expected you to translate a comparison in the grammar into an L2FC based formula.

I don't think I quite get what you mean. `l2fc(a, b) < t` translates to `(a.vaf.log2() - b.vaf.log2()) < t`, while `a < b` is meant to translate to what exactly?

johanneskoester:
> > > Since the L2FC operator always involves a comparison, we could integrate this into the Comparison path (e.g. by. adding a boolean flag field for that) or just leave it separate (as it is now, which incurs some code duplications)
> > 
> > 
> > Actually it was meant to be exactly like that. In my draft, I had implemented the L2FC path and expected you to translate a comparison in the grammar into an L2FC based formula.
> 
> I don't think I quite get what you mean. `l2fc(a, b) < t` translates to `(a.vaf.log2() - b.vaf.log2()) < t`, while `a < b` is meant to translate to what exactly?

`a < b` translates to `l2fc(a, b) < 0`.

src/utils/log2_fold_change.rs
```diff
@@ -0,0 +1,62 @@
...
+use ordered_float::NotNan;
+
+use crate::utils::comparison::ComparisonOperator;
+use crate::variants::model::AlleleFreq;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug, Clone)]
+pub(crate) struct Log2FoldChange {
+    vaf_a: AlleleFreq,
+    vaf_b: AlleleFreq,
```
johanneskoester, line 12: Why does this hold two VAFs? I would have expected just the value, in a newtype like:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Derefable)]
pub(crate) struct Log2FoldChange(#[deref] NotNan<f64>);

impl Log2FoldChange {
    /// Calculate log2 fold change of a by b (i.e. (a / b).log2()).
    /// For a == b == 0.0, returns 0.0.
    /// Returns None if fold change cannot be calculated (if both a and b are zero).
    pub(crate) fn new(a: NotNan<f64>, b: NotNan<f64>) -> Self {
        if *a == 0.0 && *b == 0.0 {
            Log2FoldChange(NotNan::new(0.0).unwrap())
        } else {
            let value = a.log2() - b.log2();
            Log2FoldChange(NotNan::new(value).expect("bug: NaN when calculating Log2FoldChange"))
        }
    }
}
```

Sorry, I had forgotten to commit this code above. Also note the special handling of the zero case.

If the VAFs are for debugging, I think it would be better to handle that outside.

tests/resources/testcases/test_cmp/scenario.yaml
```diff
@@ -0,0 +1,10 @@
+samples:
+  sample_a:
+    resolution: 100
+    universe: "[0.0,1.0]"
+  sample_b:
+    resolution: 100
+    universe: "[0.0,1.0]"
+
+events:
+  homhet: "sample_a > sample_b"
```
johanneskoester, line 10: I think your issue comes from the fact that you are not defining an "otherwise" event here. Remember, the event space has to be complete. Otherwise, the model has no other chance than accepting the frequency combinations that you allow here.

src/grammar/formula.rs
```diff
@@ -869,7 +869,7 @@ impl Formula {
             }) => Formula::Terminal(FormulaTerminal::Log2FoldChange {
                 sample_a: sample_a.into(),
                 sample_b: sample_b.into(),
-                predicate: !*predicate,
+                predicate: predicate.clone(),
```
tedil, line 872: Thanks for catching the double negation (`negate` + `apply_negations`). As clippy suggests, since l2fcp is Copy, we could just write
```suggestion
                predicate: *predicate,
```
instead

tests/lib.rs
```diff
@@ -27,6 +28,13 @@ macro_rules! testcase {
                     // Poison error can be ignored here, because it just means that the other test failed
                     // and we are safe to go on.
                     let _guard = [<$name:upper _MUTEX>].lock();
+
+                    // fern::Dispatch::new()
+                    //     .level(log::LevelFilter::Info)
+                    //     .chain(std::io::stderr())
+                    //     .apply()
+                    //     .unwrap();
+
```
tedil, line 37: Remove this?
```suggestion
```

tests/lib.rs
```diff
@@ -4,6 +4,7 @@ use std::str;
 use std::sync::Mutex;
 
 use bio::stats::{LogProb, Prob};
+use fern;
```
tedil, line 7: Remove this?
```suggestion
```

tedil (approved):

johanneskoester:
xref: https://github.com/varlociraptor/varlociraptor.github.io/pull/7

PR #203, opened 2021-08-23
test: add test for local fdr

FelixMoelder:
### Description

As title says

### QC


tests/lib.rs
```diff
@@ -193,23 +193,25 @@ fn cleanup_file(f: &str) {
...
-        )),
-        LogProb::from(Prob(alpha)),
-        local,
-    )
-    .unwrap();
+    for event_str in events {
+        varlociraptor::filtration::fdr::control_fdr(
+            &format!("{}/calls.matched.bcf", basedir),
+            Some(&output),
+            &[varlociraptor::SimpleEvent {
```
johanneskoester, line 204: The events have to be collected into this argument here instead of calling control_fdr multiple times.

johanneskoester (changes requested):

tests/lib.rs
```diff
@@ -193,16 +193,20 @@ fn cleanup_file(f: &str) {
     }
 }
 
-fn control_fdr(test: &str, event_str: &str, alpha: f64, local: bool) {
+fn control_fdr(test: &str, events: Vec<&str>, alpha: f64, local: bool) {
```
johanneskoester, line 196:
```suggestion
fn control_fdr(test: &str, events: &[&str], alpha: f64, local: bool) {
```

Take a slice here (this enables to either use Vec or Array below. Best practice: always make function params as generic as possible.

tests/lib.rs
```diff
@@ -229,44 +233,66 @@ fn assert_call_number(test: &str, expected_calls: usize) {
 
 #[test]
 fn test_fdr_control1() {
-    control_fdr("test_fdr_ev_1", "SOMATIC", 0.05, false);
+    control_fdr("test_fdr_ev_1", vec!["SOMATIC"], 0.05, false);
```
johanneskoester, line 236:
```suggestion
    control_fdr("test_fdr_ev_1", ["SOMATIC"], 0.05, false);
```
and analogous for the tests below.

tests/lib.rs
```diff
@@ -229,44 +233,66 @@ fn assert_call_number(test: &str, expected_calls: usize) {
...
+#[test]
+fn test_fdr_control_local3() {
+    control_fdr(
+        "test_fdr_local3",
+        vec![
+            "GERMLINE",
+            "SOMATIC_TUMOR_HIGH",
+            "SOMATIC_TUMOR_LOW",
+            "ARTIFACT",
+            "FFPE_ARTIFACT",
```
johanneskoester, line 280: Why do you now give all the events here? Just GERMLINE and SOMATIC_TUMOR_LOW, like in the config.

johanneskoester (changes requested):

PR #211, opened 2021-09-09
feat: improved, more efficient, bias computation

johanneskoester:
### Description

Bias events are now calculated independently per sample, contamination is not considered, a flat prior is used.

### QC


johanneskoester:
This leads to less interpretable results, and also can lead to FPs on the CHM13 dataset. Hence, we keep the current, more conservative approach.

PR #226, opened 2021-10-19
feat: better homopolymer error detection

johanneskoester:
### Description

More specific detection of homopolymer errors by estimating the multinomial distribution of such errors from the given alignments.

### QC


src/estimation/alignment_properties.rs
```diff
@@ -13,18 +15,120 @@ use ordered_float::NotNan;
...
     pub(crate) frac_max_softclip: Option<f64>,
     pub(crate) max_read_len: u32,
+    #[serde(default = "default_homopolymer_error_model")]
+    pub(crate) homopolymer_error_model: HashMap<i8, f64>,
     #[serde(default)]
     initial: bool,
 }
 
 impl AlignmentProperties {
+    pub(crate) fn update_homopolymer_error_model(
```
johanneskoester, line 50: @tedil can you please have a close look at this and check whether the error ratest estimated here would make sense as parameters for the homopolypair-hmm?

tedil: So the current homopolypair-hmm implementation assumes that extending a homopolymer (e.g. from 2,1 to 3,1 to 4,1 etc) always happens at **roughly** the same rate (which is what I observed in some preliminary experiments on GIAB data) [there's only one homopolymer state for each base {A/T homopolymers are more likely than C/G ones if I recall correctly} and for target/query, not multiple sequential states, though we could easily change that I guess].
i.e. there's one initial homopolymer-open probability akin to gap-open and one homopolymer-extend probability akin to gap-extend.

From what I recall, the extend probability is, depending on base+target/query, around 0.04 to 0.18 or so, so at least the 1 → 2 step is fine. Not sure about the initial probabilities atm. Will report back later.

johanneskoester: Good point. It would be fine for me if you extended the homopolymer error model estimation in the alignment properties to be base specific if the rest fits.

PR #227, opened 2021-10-20
feat!: Introduce homopolymer aware pair HMM

tedil:
### Description

This PR includes the homopolymer aware pair HMM from rust-bio for handling reads from oxford nanopore sequencing.
There's one breaking change: the error rates are now specified in a yaml file instead of separate cli options. This makes it also possible to have sequencing machine specific error rates easily available from common resources.

### QC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### [4.11.4](https://www.github.com/varlociraptor/varlociraptor/compare/v4.11.3...v4.11.4) (2022-03-31)
+
+
+### Bug Fixes
+
+* only check for homopolymer errors if the indel operation is alone in the variant ([#267](https://www.github.com/varlociraptor/varlociraptor/issues/267)) ([61aa7ba](https://www.github.com/varlociraptor/varlociraptor/commit/61aa7ba5a60cd80c5ac21af7cea1e94b91c79406))
+
 ### [4.11.3](https://www.github.com/varlociraptor/varlociraptor/compare/v4.11.2...v4.11.3) (2022-02-24)
 
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2308,7 +2308,7 @@ dependencies = [
 
 [[package]]
 name = "varlociraptor"
-version = "4.11.3"
+version = "4.11.4"
 dependencies = [
  "GSL",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "varlociraptor"
-version = "4.11.3"
+version = "4.11.4"
 authors = ["Johannes Köster <johannes.koester@tu-dortmund.de>", "David Lähnemann <david.laehnemann@uni-duesseldorf.de>"]
 description = "A library for calling of genomic variants using a latent variable model."
 repository = "https://github.com/varlociraptor/varlociraptor"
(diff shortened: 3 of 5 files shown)


johanneskoester (approved):
Very nice! Can you add a test case with nanopore data?

tedil:
I think this branch is finally ready for review. This involves quite some changes, some breaking the CLI:

1. alignment property estimation has changed: it now counts matches, gap and hop (homopolymer) indels from the CIGAR representation, where the hops are counted per base, respecting case (i.e. lowercase bases usually mean "masked in the reference" for example for repeat regions). This also means that the alignment property json file now has additional fields. And it's pretty printed!
2. spurious ins/del parameters are now derived from the gap indel counts, instead of being a cli parameter. The old default values are still used as a fallback
3. homopolymer extension probabilities are likewise calculated from the hop counts.
4. there is a new pairhmm-mode available (which is the point of this whole PR): `--pairhmm-mode homopolymer`, which realigns using the homopolypairhmm instead of the default pairhmm (`exact`) or pathhmm (`fast`). This is considerably slower, but models homopolymer errors as happen in nanopore data. (There is still a question of whether homopolymer deletions exhibit a secondary effect, see homopolymer_counts
If so, deletions might need additional states to allow some kind of mixture model)
5. the pairhmm override in the testcases was not in effect, not sure if that was intentional or not

src/cli.rs
```diff
@@ -14,6 +14,7 @@ use bio::io::fasta;
 use bio::stats::bayesian::bayes_factors::evidence::KassRaftery;
 use bio::stats::{LogProb, Prob};
 use itertools::Itertools;
+use rust_htslib::bam::Read;
```
johanneskoester, line 17: Seems to be unused according to clippy?

src/cli.rs
```diff
@@ -269,15 +245,17 @@ pub enum PreprocessKind {
...
             default_value = "exact",
-            help = "PairHMM computation mode (either fast or exact). Fast mode means that only the best \
+            help = "PairHMM computation mode (either fast, exact or homopolymer). Fast mode means that only the best \
                     alignment path is considered for probability calculation. In rare cases, this can lead \
                     to wrong results for single reads. Hence, we advice to not use it when \
                     discrete allele frequences are of interest (0.5, 1.0). For continuous \
                     allele frequencies, fast mode should cause almost no deviations from the \
                     exact results. Also, if per sample allele frequencies are irrelevant (e.g. \
-                    in large cohorts), fast mode can be safely used."
+                    in large cohorts), fast mode can be safely used. \
```
johanneskoester, line 256:
```suggestion
                    in large cohorts), fast mode can be safely used. \
                    Note that fast and exact mode are not suitable for ONT data.\
```

src/cli.rs
```diff
@@ -710,68 +681,91 @@ pub fn run(opt: Varlociraptor) -> Result<()> {
...
+                                    .realigner(realignment::PairHMMRealigner::new(
+                                        reference_buffer,
+                                        gap_params,
+                                        realignment_window,
+                                    ))
+                                    .build();
+                            processor.process()?;
+                        }
+                        _ => panic!("Unknown pairhmm mode '{}'", pairhmm_mode),
+                    };
```
johanneskoester, line 768: Looks like we should now first init the processor, and only add the mode specific args in the match statement.

johanneskoester (changes requested):
Great work! THere are some clippy warnings in addition to those that I have commented on.

src/cli.rs
tedil: Yes, I tried simplifying that but had some type issues (which aren't even easily solved by box-ing), so I left it as is for the time being.

johanneskoester (approved):

PR #244, opened 2021-12-14
feat: optional uniform prior for non absent variants in the scenario

johanneskoester:
### QC


johanneskoester:
We have a better solution now!

PR #262, opened 2022-02-23
fix: strum versioning

tedil:
### Description

This PR sets strum version to 0.23.0 which complies with rust-bio's version pinning to < 0.24; this also replaces strum_macros with strum's "derive" feature.

### QC


tedil:
Closed because this conflates commit histories, will do cleaner PRs.

Issue #261, opened 2022-02-23
Installation Failure. error[E0277]


bwubb:
Hello,

  I am unable to install/upgrade to version 4.11.2. The error is completely strange to me so I am not sure what I should begin troubleshooting.

```
...
Compiling varlociraptor v4.11.2
error[E0277]: the trait bound `u128: bio::pattern_matching::myers::BitVec` is not satisfied
  --> /home/bwubb/.cargo/registry/src/github.com-1ecc6299db9ec823/varlociraptor-4.11.2/src/variants/evidence/realignment/edit_distance.rs:19:11
   |
19 |     Short(myers::Myers<u128>),
   |           ^^^^^^^^^^^^^^^^^^ the trait `bio::pattern_matching::myers::BitVec` is not implemented for `u128`
   |
note: required by a bound in `bio::pattern_matching::myers::Myers`
  --> /home/bwubb/.cargo/registry/src/github.com-1ecc6299db9ec823/bio-0.39.2/src/pattern_matching/myers/simple.rs:17:8
   |
17 |     T: BitVec,
   |        ^^^^^^ required by this bound in `bio::pattern_matching::myers::Myers`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `varlociraptor` due to previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `varlociraptor v4.11.2`, intermediate artifacts can be found at `/tmp/cargo-installrzquek`

Caused by:
  build failed
```
Any help is greatly appreciated. Thank you

-bwubb

johanneskoester:
We have seen this as well recently, seems to be some incompatibility between dependencies. Note that you can always install the latest Varlociraptor via Bioconda though.

bwubb:
Thank you for the response. I have deployed the conda solution, which of course works. My IT admins frown on conda for whatever reason, but oh well.

johanneskoester:
Fixed in release 4.11.3. Thanks for reporting!

PR #264, opened 2022-02-24
fix: check if cfg has_u128 or not

tedil:
### Description

With rust-bio 0.39.2, the BitVec trait has `#[cfg(has_u128)]` checks, so we also do those checks.

### QC


tedil:
Ah, this doesn't work because we rely on the `u128` for pattern length.

tedil:
This should fix #261

johanneskoester (approved):
Thanks a lot, perfect, as always!

PR #269, opened 2022-04-11
test: somatic normal prior tests

dlaehnemann:
### Description

This PR adds two test cases, one with only very slight evidence for a somatic mutation in the normal sample (should be something like two reads) and one with very clear evidence for a somatic mutation in the normal sample. For both cases, we test with two scenarios each: once without a somatic mutation rate prior specified (thus defaulting this prior to `0.0`) for the normal sample and once with `somatic-effective-mutation-rate: 1e-6` specified for the normal sample.

### QC


dlaehnemann:
Try reopening.

PR #274, opened 2022-05-03
fix: convert empty VAF spectrum to false and ensure that an empty spectrum evaluates as zero probability

johanneskoester:
### QC


dlaehnemann (approved):
Looks good. Thanks for the quick fixes!

PR #287, opened 2022-06-13
chore: Make UnsortedVariantFile error more informative

tedil:
### Description

This PR includes the two loci which violate the `previous_locus <= current_locus` condition in the error.

### QC

### DIFF
--- a/src/utils/variant_buffer.rs
+++ b/src/utils/variant_buffer.rs
@@ -77,7 +77,11 @@ impl VariantBuffer {
                                 && locus.pos() < previous_locus.pos()
                             {
                                 // unsorted input file, fail with an error
-                                return Err(errors::Error::UnsortedVariantFile.into());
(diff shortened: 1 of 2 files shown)


src/utils/variant_buffer.rs
```diff
@@ -77,7 +77,11 @@ impl VariantBuffer {
                                 && locus.pos() < previous_locus.pos()
                             {
                                 // unsorted input file, fail with an error
-                                return Err(errors::Error::UnsortedVariantFile.into());
+                                return Err(errors::Error::UnsortedVariantFile {
+                                    previous_locus: previous_locus.clone(),
+                                    current_locus: locus.clone(),
```
johanneskoester, line 82:
```suggestion
                                    current_locus: locus,
```

johanneskoester (changes requested):

PR #293, opened 2022-09-27
fix: Adhere to number specifications from VCF v4.3

tedil:
### Description

This PR adjusts the `Number=X` information of the VCF header written in the preprocessing step such that it matches the VCF specification v4.3.

### QC


johanneskoester (approved):
Ok. I don't like all of them, but better follow the spec.

tedil:
> Ok. I don't like all of them, but better follow the spec.

Me neither.

PR #310, opened 2022-12-01
fix: feed reported allele freq dist from all VAF estimates, not only the MAP.

johanneskoester:
### Description

TODO: this does not yet work, as it creates many very similar VAFs for which probs jump around
(because the model does not always use exactly the same grid due to adaptive integration).

Example (rounded vafs): 

0.44=140.19,0.45=15.47,0.45=-13.82,0.45=1.84,0.45=-13.84,0.45=140.09

Certain VAFs are simply not evaluated for the MAP.

### QC


johanneskoester:
This does not work as easily, we would need to somehow integrate over a wider area around the VAFs of other samples. For now, we can also close this.

PR #319, opened 2022-12-29
feat: add experimental support for the estimation of contaminations

johanneskoester:
This has been validated on a few samples, but still needs a proper study (which is currently happening)

### QC


johanneskoester:
We will skip tests and docs this time and for now release this as experimental feature for internal use (in evaluation workflows).

PR #371, opened 2023-05-11
fix: Improved parameter estimation for (H)PHMM

tedil:
### Description

This PR updates the (H)PHMM parameter estimation routine:
 - counts transitions a little differently then before: 
   1. conceptually expand cigar strings (i.e. `3M2D1I2M` → `1M1M1M1D1D1I1M1M`)
   2. iterate consecutive pairs e.g. `(1M, 1M)` and look at rseq and qseq at the corresponding positions
   3. classify that into any of the 82 transitions, e.g. `(1M, 1M)` together with `((T, T), (G, G))` corresponds to a `match T → match G` transition
 - this will only look at `N` records from the SAM/BAM/CRAM, where `N` is a number calculated as a multinomial sample size estimation as described in [https://www.jstor.org/stable/2683352](https://www.jstor.org/stable/2683352)
When the number of alignments in the file is known (which is the case for indexed BAM files but not for indexed SAM/CRAM), can use a finite population correction version of the estimate instead.
 - currently, iterates through the whole file with a fixed step size. However, this isn't any faster than simply looking at every record, so do any combination of the following instead:
   1. only use the *first* `N` records (possibly biased subsample, but definitely quicker)
   2. for CRAM, set lowlevel htslib reading options to skip fields that aren't used for estimation purposes (potentially quicker, but only for CRAM?)
   3. look at samtools subsample or simply use that as a preprocessing step
   4. make better use of indices for random access

### QC


tedil:
`bam_utils.rs` can be replaced if https://github.com/rust-bio/rust-htslib/pull/387 is integrated

PR #512, opened 2025-06-26
CNV alignment properties

Addimator:
### Description

This pull request adds support for calculating and storing average depth and average depth per CG ratio in the alignment properties. These metrics are computed based only on genomic regions that are not affected by copy number variations (CNVs). To enable this, users can optionally provide a BCF file containing CNV candidate regions. 

### QC


Addimator:
@coderabbitai review

PR #536, opened 2026-01-29
fix: sort samples lexicographically

Addimator:
### Description

This PR ensures that all samples of an event are sorted lexicographically before evaluation. As a result, Varlociraptor now produces deterministic results for events involving multiple samples.

Previously, the evaluation order of samples depended on their original ordering. Because the adaptive integration is recursive, the sample that was examined first influenced the integration path. This could lead to slightly different numerical results for logically identical events whose samples were listed in a different order.

For example, consider the following event:

`present: "sample1 == sample2 & sample1:]0.0,1.0] & sample2:]0.0,1.0]"`

In this case, it makes a difference whether the adaptive integration starts with sample1 and evaluates sample2 only at the support points determined by sample1 (due to the L2FC dependency), or vice versa. Depending on the order, the integration explored different regions first and produced slightly different results.

By enforcing a consistent, lexicographical ordering of samples, this PR eliminates this source of nondeterminism and guarantees reproducible results regardless of how samples are specified in the event.

### QC

* [x] Run benchmark (check this box to trigger the benchmark workflow)

## Summary by CodeRabbit

* **New Features**
  * Normalization now augments formulas with any missing per-sample terms and integrates them into final normalized expressions.

* **Bug Fixes**
  * Operand sorting improved with multi-criteria ordering for consistent results.
  * VAF/observability range handling fixed to respect left/right exclusivity with epsilon adjustments.
  * Spectrum split logic corrected for exclusivity on singleton ranges.

* **Tests**
  * Added three end-to-end tests and fixtures for single- and two-sample sorting scenarios.


src/grammar/formula.rs
```diff
@@ -458,16 +458,91 @@ impl Formula {
...
+                operands.sort();
+                operands.sort_by(|a, b| {
+                    let key = |f: &Formula| match f {
+                        Formula::Terminal(FormulaTerminal::Log2FoldChange { sample_a, .. }) => {
+                            (0, sample_a.clone())
+                        }
+                        Formula::Terminal(FormulaTerminal::Atom { sample, .. }) => {
+                            (1, sample.clone())
+                        }
+                        _ => (2, String::new()),
```
johanneskoester: I think this should work without the clone. Use an Option of a ref to string as the second item.
```suggestion
                            (0, Some(sample_a))
                        }
                        Formula::Terminal(FormulaTerminal::Atom { sample, .. }) => {
                            (1, Some(sample))
                        }
                        _ => (2, None),
```
And finally, I think you can stick to the previous way of sorting by key and just return those keys instead of having to run the cmp below.

src/grammar/formula.rs
```diff
@@ -458,16 +458,91 @@ impl Formula {
...
+    fn compute_included_samples(&self, included: &mut HashSet<String>) {
+        match self {
+            Formula::Conjunction { operands } | Formula::Disjunction { operands } => {
+                for operand in operands {
+                    operand.compute_included_samples(included);
+                }
+            }
+
+            Formula::Terminal(FormulaTerminal::Atom { sample, .. }) => {
+                included.insert(sample.clone());
```
johanneskoester: Can we avoid the clone here and put references into the hashset?

src/grammar/formula.rs
```diff
@@ -458,16 +458,91 @@ impl Formula {
...
+            }
+
+            _ => {}
+        }
+    }
+
+    /// Add missing samples to the formula based on the given scenario.
+    pub fn add_missing_samples(&mut self, scenario: &Scenario, contig: &str) -> Result<()> {
```
johanneskoester: If I am not wrong, such a completion function existed before, just not at this stage, but later when samples are converted to integer ids or when they the formula is converted into a VAFTree. Am I right? So, two questions:
1. Can't we do the sorting after that operation?
2. If not, the code over there for adding missing samples should be removed instead.
I would however prefer the former.

johanneskoester (changes requested):

src/grammar/formula.rs
Addimator: Yeah, originally the missing samples have been added during conversion to the VAFtree. The problem is here, that the samples have no name at this stage and therefore can't be sorted lexicographically. 

I can remove the redundant code in the VAFtree though.

johanneskoester: But they can be sorted via their id, which is just a usize!

johanneskoester: That's still deterministic and even faster. I would prefer that as it limits the changes to the code base.

Addimator: Yes, you are completely right! I did not see that scenario.samples is a BTreeMap already sorted by name! So I will have to look into the code there and think about a nice strategy to reorder the nodes of the tree according to l2fc and sample_idx.

src/grammar/formula.rs
Addimator: If I understand it correctly I need to clone the Strings because of lifetime issues. I can use sort_by_key though:
https://stackoverflow.com/questions/68390443/how-to-sort-a-veccowstr-cowstr-by-key-without-cloning

src/grammar/formula.rs
```diff
@@ -458,28 +458,232 @@ impl Formula {
...
+                        // 2️⃣ neuen Conjunction bauen
+                        let mut conj_ops = Vec::with_capacity(new_ops.len() + 1);
+                        conj_ops.append(&mut new_ops);
+
+                        // 3️⃣ zurückschreiben an exakt dieselbe Stelle
+                        *op = Formula::Conjunction { operands: conj_ops };
+                    }
+
+                    // println!("Operands after processing child in disjunction: {op}");
+                }
```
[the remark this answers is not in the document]
Addimator: No, this is the correct approach. Each disjunction starts a new term, which must include all samples, so a fresh seen set is required for each term.

The situation in your example cannot occur because we always call expand_expressions() beforehand. After expansion, the formula would look like this:

`(sample1:]0,1] & sample2:]0,0.5]) | (sample1:]0,1] & sample2:]0.5,1])`

This ensures that each term is fully self-contained with all relevant samples.

src/grammar/formula.rs
```diff
@@ -461,15 +461,109 @@ impl Formula {
...
+                // missing samples. This ensures we only inject once.
+                if !is_last {
+                    return Ok(None);
+                }
+                let mut missing_terms = Vec::new();
+                for (name, sample_data) in scenario.samples() {
+                    if seen.contains(name) {
+                        continue;
+                    }
+                    seen.insert(name.to_string());
```
johanneskoester:
```suggestion
                    seen.insert(name.to_owned());
```

johanneskoester (changes requested):

johanneskoester (approved):

Issue #158, opened 2021-05-04
Support genotyping with the grammar


nh13:
It would really useful to have as part of the grammar the ability to call genotypes, along with accurate allele depth and frequency for those genotypes.  A motivating example would be [tumor-normal somatic calling discussed in another issue](https://github.com/varlociraptor/varlociraptor/issues/153#issuecomment-829138510).  I could see the grammar allow defining a mapping from the probability of an event (or the sum from multiple independent events) to genotype.  The genotype would be the event with the maximum probability.  Allele depths and frequency can be calculated from the observations (eg. see this [comment](https://github.com/varlociraptor/varlociraptor/issues/153#issuecomment-829453913)).  One could even construct likelihoods (FROMAT/PL tag) from the posteriors (I think the posteriors as computed are the FORMAT/PP tag).  The genotyping would need to handle [multiple alleles](https://github.com/varlociraptor/varlociraptor/issues/157) and combinations of variant types (eg. SNV on one haplotype, indels on the other)

PR #558, opened 2026-07-30
feat(grammar): verify pairwise event disjointness at scenario compilation

Adelagric:
## Motivation
The docs require events to be pairwise disjoint — no combination of allele
frequencies may satisfy more than one event, since the generic model treats
events as mutually exclusive when summing posteriors. This was only enforced by
a heuristic in `Scenario::validate` that flagged a pair `(e1, e2)` when
`e1 | e2` happened to equal a third *named* event (with a `TODO` noting it was
unreliable). It misses the documented counter-example: `germline` = `normal:0.5`
overlapping a `somatic_normal` = `normal:]0.0,0.5]`.

## Change
Replace it with an exact per-contig pairwise check over `VAFTree` clauses
(root-to-leaf paths). Two events overlap when some clause pair shares the same
variant and log2-fold-change literals and has a non-empty per-sample VAF
intersection within the declared universe. The `OverlappingEvents` error now
names the offending pair and exhibits a concrete VAF witness, e.g.:

    the following events are not disjunct: "germline" and "somatic_normal" (witness: normal=0.5)

Boolean-terminal handling uses exact set matching: sound (no false positives),
consistent with the model's own `VAFTree::contains` semantics, so events
distinguished only by log2-fold-change predicates or variant terminals stay
disjoint. The check is exact and complete on pure allele-frequency events.

## Tests
- New unit tests: documented counter-example, disjoint tumor/normal, identical
  events, interior-range witness, full overlap fixture, l2fc-disjoint,
  variant-disjoint.
- Integration: `test_overlapping_events` still panics; `test_l2fc`, `test_cmp`,
  `test40/55/57/59` still pass.

## Also
Fixes a latent overlap this check surfaced in
`tests/resources/prior/scenarios/pedigree.scenario.yaml`: `denovo_child` and
`denovo_sibling` were both true at `mother=0, father=0, child=0.5, sibling=0.5`
(each left the other child unconstrained). Each de novo event now requires the
other child to be absent — also the correct semantics.

## Summary by CodeRabbit

* **New Features**
  * Improved detection of overlapping genetic events, including compatible variant constraints.
  * Error messages now identify conflicting events and provide concrete example assignments.
  * Added support for frequency-spectrum intersections and representative allele-frequency values.

* **Bug Fixes**
  * Improved scenario validation using sample-specific frequency constraints.
  * Corrected pedigree event expressions to include required cross-individual constraints.
  * Updated a brain event scenario to exclude the midpoint value from matching.

* **Tests**
  * Added regression coverage for overlap detection, witness details, edge-case frequency ranges, and example scenarios.

### DIFF
--- a/src/grammar/formula.rs
+++ b/src/grammar/formula.rs
@@ -1052,6 +1052,47 @@ impl VAFSpectrum {
             VAFSpectrum::Range(range) => range.is_complete(),
         }
     }
+
+    /// Intersection of two spectra. Returns an empty spectrum if they are disjoint.
+    pub(crate) fn intersect(&self, other: &VAFSpectrum) -> VAFSpectrum {
+        match (self, other) {
+            (VAFSpectrum::Range(a), VAFSpectrum::Range(b)) => {
+                let r = a.intersect(b);
+                if r.is_empty() {
+                    VAFSpectrum::empty()
+                } else if r.is_singleton() {
+                    VAFSpectrum::singleton(r.start)
+                } else {
+                    VAFSpectrum::Range(r)
+                }
+            }
+            (VAFSpectrum::Range(r), VAFSpectrum::Set(s))
+            | (VAFSpectrum::Set(s), VAFSpectrum::Range(r)) => {
+                VAFSpectrum::Set(s.iter().filter(|v| r.contains(**v)).cloned().collect())
+            }
+            (VAFSpectrum::Set(a), VAFSpectrum::Set(b)) => {
+                VAFSpectrum::Set(a.intersection(b).cloned().collect())
+            }
+        }
+    }
+
+    /// Pick a representative allele frequency that is contained in this spectrum, or `None` if the
+    /// spectrum is empty. For ranges, the midpoint is returned, which is strictly interior and thus
+    /// contained regardless of the range's exclusivity.
+    pub(crate) fn representative(&self) -> Option<AlleleFreq> {
+        match self {
+            VAFSpectrum::Set(s) => s.iter().next().copied(),
+            VAFSpectrum::Range(r) => {
+                if r.is_empty() {
+                    None
+                } else if r.is_singleton() {
+                    Some(r.start)
+                } else {
+                    Some(AlleleFreq((*r.start + *r.end) / 2.0))
+                }
+            }
+        }
+    }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, TypedBuilder, Hash, CopyGetters)]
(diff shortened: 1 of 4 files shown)


Adelagric:
Hi @johanneskoester — gentle ping on this one.

Quick recap: it replaces the heuristic in `Scenario::validate` (the one with the TODO) with an actual per-contig pairwise disjointness check over the VAF-tree clauses, and reports a concrete VAF witness when two events overlap. While doing that it turned up a real overlap in `tests/resources/prior/scenarios/pedigree.scenario.yaml` (`denovo_child` / `denovo_sibling` are both true at mother=0, father=0, child=0.5, sibling=0.5); the second commit fixes the fixture.

Still applies cleanly on current master (#557). Happy to adjust the scope or split the fixture fix out if you'd prefer.

Adelagric:
Pushed a third commit that compiles every example scenario under `tests/resources/prior/scenarios` — it fails on the old pedigree fixture with the same pair and witness as above, so the invariant is now guarded there too.

Adelagric:
Two follow-ups after CodeRabbit's pass, both pushed:

- The variant-terminal comparison is now a real compatibility check rather than exact set equality. A locus is in one of thirteen contexts (the twelve concrete SNVs, or no SNV), and a terminal holds in a context under the same IUPAC matching as `GenericPosterior::density`; two clauses are compatible iff some context satisfies all terminals of both. So `C>T & tumor:[0.1,0.2]` vs `tumor:[0.1,0.2]` is now rejected (witness names `variant C>T`), while `C>T` vs `G>A`, or a terminal vs its negation, stay disjoint. `VAFSpectrum::representative` also validates the midpoint against the range now.
- Running the full integration suite locally turned up `test69`: its `denovo` and `mosaik_brain` events are both true at brain=0.5, so the scenario fails the new validation. I excluded 0.5 from the mosaic event; the testcase's expected allele frequencies are unchanged. The CI workflows on this PR are waiting for approval, which is why this did not show up before.

Locally: `cargo test` 31 unit + 209 integration green.

Adelagric:
On CodeRabbit's last point (checking declared events against the implicit `absent` event): the observation is right — `calling.rs` registers `absent` next to the declared events and rust-bio normalises by the sum over the whole universe, so an event that admits the all-zero VAF point (e.g. `hom_ref: "tumor:0.0"`) has that mass counted twice in the marginal, and both `PROB_ABSENT` and `PROB_HOM_REF` come out as `p/(2p+rest)`.

I tried enforcing it and four existing testcases declare exactly such events (`pattern_too_long`, `test37`, `test70`, `test_prinz_pacbio_zero`) — genotyping-style scenarios like the ones #158 asks for. Rejecting them is a user-visible change, and the fix might rather belong on the model side (skip the implicit `absent` when a declared event covers it), so I left it out of this PR. Happy to do either as a follow-up if you tell me which way you'd like it.

johanneskoester (approved):
Excellent work! Thanks a lot!

